PR #6, opened 2015-04-30
Updates for psc 0.7

hdgarrood:
I still need to test the `ordArray` instance, it's not immediately clear how to do that (since if you try to install `purescript-console`, you get another version of `purescript-prelude` in your bower_components...)

### DIFF
--- a/src/Prelude.purs
+++ b/src/Prelude.purs
@@ -5,7 +5,6 @@ module Prelude
   , const
   , asTypeOf
   , otherwise
-  , (:), cons
   , Semigroupoid, compose, (<<<), (>>>)
   , Category, id
   , Functor, map, (<$>), (<#>), void
@@ -115,30 +114,6 @@ asTypeOf x _ = x
 -- | ```
 otherwise :: Boolean
 otherwise = true
-
--- | Attaches an element to the front of an array, creating a new array.
--- |
--- | ```purescript
--- | cons 1 [2, 3, 4] = [1, 2, 3, 4]
--- | ```
--- |
--- | Note, the running time of this function is `O(n)`.
-foreign import cons
-  """
-  function cons(e) {
-    return function(l) {
-      return [e].concat(l);
-    };
-  }
-  """ :: forall a. a -> [a] -> [a]
-
-infixr 6 :
-
--- | An infix alias for `cons`.
--- |
--- | Note, the running time of this function is `O(n)`.
-(:) :: forall a. a -> [a] -> [a]
-(:) = cons
 
 infixr 9 >>>
 infixr 9 <<<
--- a/src/Prelude.purs
+++ b/src/Prelude.purs
@@ -174,7 +174,7 @@ class Functor f where
 instance functorFn :: Functor ((->) r) where
   map = compose
 
-instance functorArray :: Functor [] where
+instance functorArray :: Functor Array where
   map = arrayMap
 
 foreign import arrayMap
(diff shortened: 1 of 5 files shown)


src/Prelude.purs
```diff
@@ -811,13 +786,43 @@ instance ordChar :: Ord Char where
...
+
+          if (result === null) {
+             result = eq
+           }
+          return result
+        }
+      }
+    }
+  }
+  """ :: forall a. (a -> a -> (Ordering -> Unit) -> Unit) -> Ordering -> Array a -> Array a -> Ordering
```
paf31: I ran into the same issues when I tried to implement this. I landed on using `Int` as the return type (-1, 0, 1), which isn't ideal, but I think it's better than paying the price of four extra function arguments. I'd also rather avoid the impure callback.

hdgarrood:
The `ordArray` instance is broken - `compare [] [0]` gives `EQ`. I'm working on tests now.

paf31:
Looks nice thanks. For the testing, I usually delete the extra Prelude from `bower_components`. I was thinking we would add a separate directory with its own `bower.json` for tests anyway.

src/Prelude.purs
paf31: See the version in `tests/prelude.purs` in the compiler repo. It uses `Number`, but you get the gist.

garyb: Wouldn't the extra arguments be simpler? That doesn't impose any cost on using the function as they're curried in at instance-definition-time.

paf31: I guess that's true. I maintain my stance on the impure bit though :)

garyb: Agreed! I'm playing around with an alternative implementation now.

garyb: There's a bug in the prelude version by the way, both inner functions accept `ys`.

garyb: Also `i` is never incremented!

paf31: Oops :) I guess it's not used then.

paf31: D'oh. I actually put it in there to make things compile, then forgot about it. I meant to go back and fix it up.

hdgarrood:
There's a new `ordArray` instance in here now, which seems to work. It's based on the one in the compiler currently.

You can run the tests by doing:

```
$ psc test/Main.purs src/**/*.purs bower_components/*/src/**/*.purs --main | node
```

I'm not good enough with gulp to work out how to make this into `gulp test`.

garyb:
I did something similar in `purescript-arrays`: https://github.com/purescript/purescript-arrays/blob/master/gulpfile.js#L44-L49

paf31:
Looks good to me. I was thinking that for tests, we could eventually work on a separate test suite project for each core library. We could even use `quickcheck` that way.

@garyb Looks good, merge?

garyb:
Maybe we should break out a `purescript-assert` or something that provides zero-dependency assertions so we can use it across any of the projects above quickcheck in the dependency hierarchy.

garyb:
Yep, :shipit:

paf31:
:+1: for `purescript-assert`, but not for the reason you said necessarily. I would rather have the option of using `quickcheck`, but separate the project out so we don't get circular dependencies.

garyb:
My only concern about that is the thing we have with the `core-tests` library currently - nobody knows it exists, and we don't check those tests when making updates or add new cases for new features. That's not unfixable, we just have to make sure whatever it is is integrated into the travis build for the corresponding libraries.

PR #13, opened 2015-05-01
Add more lattice laws

### LINKED ISSUE #9: Lattice laws
Based on https://en.wikipedia.org/wiki/Lattice_(order) should `Lattice` have the following additional laws?
- Supremum:
  - `x || y >= x`
  - `x || y >= y`
- Infimum:
  - `x && y <= x`
  - `x && y <= y`

We could potentially go even further and require that `x || y` is the least upper bound of `{x, y}`, and similar for `&&`.

hdgarrood:
Fixes #9


paf31:
:+1: Thanks!

PR #17, opened 2015-05-02
Raised precedence of `(#)`.

### LINKED ISSUE #2: Raise precedence of (#)
This has been something that has bothered me for too long.

Right now `(#)` and `($)` have the same precedence level, so you cannot use them both inline without resorting to parens. Could we make it `infixl 1 #`?

joneshf:
Couldn't run tests to make sure it works properly, so merge with caution.

Should close #2.

### DIFF
--- a/docs/Prelude.md
+++ b/docs/Prelude.md
@@ -1013,6 +1013,13 @@ instance boundedOrdering :: Bounded Ordering
 ```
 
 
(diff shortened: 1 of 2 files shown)


garyb:
:+1:

joneshf:
Thanks!

PR #22, opened 2015-05-13
Don't export Unit data constructor

hdgarrood:
Exporting the Unit data constructor is a little confusing, and I don't see any harm in removing it from the export list.

I wasn't able to run the tests :( this currently depends on an unreleased version of gulp-purescript, and I couldn't get `npm link` to work.

### DIFF
--- a/docs/Prelude.md
+++ b/docs/Prelude.md
@@ -4,7 +4,6 @@
 
 ``` purescript
 newtype Unit
-  = Unit {  }
 ```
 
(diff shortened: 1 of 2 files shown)


garyb:
:+1: 

To use the new `gulp-purescript` you need to check it out locally, use `npm run build`, then `npm link`, then in `purescript-prelude` use `npm link gulp-purescript && npm i`.

PR #25, opened 2015-05-24
Update docs for `Ring` (fixes #24)

### LINKED ISSUE #24: Ring law is unclear
We currently have:

> Additive inverse: `a + (-a) = (-a) + a = zero`

However, this is a little unclear, since there is no unary prefix `(-)` available. I think we should do one of the following:
- Swap `sub` and `negate` around, so that `negate` is part of `Ring`, and `(-)` is defined as `x - y = x + negate y`.
- Reformulate the additive inverse law in terms of binary `(-)`, for example: `a - a = (zero - a) + a = zero`?

hdgarrood:


michaelficarra:
LGTM

paf31:
:+1:

PR #26, opened 2015-05-24
Add Boolean Semiring instance

hdgarrood:
Credit for noticing this goes to @michaelficarra.

### DIFF
--- a/docs/Prelude.md
+++ b/docs/Prelude.md
@@ -515,6 +515,7 @@ class Semiring a where
 ``` purescript
 instance semiringInt :: Semiring Int
 instance semiringNumber :: Semiring Number
+instance semiringBoolean :: Semiring Boolean
 instance semiringUnit :: Semiring Unit
 ```
 
@@ -975,6 +976,5 @@ a human-readable `String` representation.
 While not required, it is recommended that for any expression `x`, the
 string `show x` be executable PureScript code which evaluates to the same
 value as the expression `x`.
-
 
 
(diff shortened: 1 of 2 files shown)


garyb:
Could make it an instance for values with a `BoundedLattice` instance generally perhaps? Not sure if that might be dodgy in regard to overlaps though.

hdgarrood:
Maybe a newtype in the spirit of `Multiplicative` or `Additive` would make sense?

``` purescript
newtype SupInf a = SupInf a
instance semiringSupInf :: BoundedLattice a => Semiring (SupInf a)
```

edit: with a better name, perhaps

michaelficarra:
:grin: I helped!

garyb:
@hdgarrood that would make sense, perhaps belongs in `purescript-semirings` then. Indeed I don't know what the name would be though :wink:

hdgarrood:
`LatticeSemiring` perhaps? Also I guess there wouldn't be a need for a `Boolean Semiring` instance if we had that?

garyb:
Yeah, that's what I'd suggest - we just have the newtype'd version.

hdgarrood:
:+1:

paf31:
Doesn't every lattice have two semirings?

hdgarrood:
I'm not sure. Is the other one you're thinking of obtained by flipping the order (like Haskell's `Down`) and swapping `(+)` with `(*)`?

paf31:
Yes, I think so.

hdgarrood:
Maybe we could do that as `instance Semiring a => Semiring (Down a) where [...]`?

garyb:
:+1:

PR #27, opened 2015-05-27
Some extra functions, fixes #20

### LINKED ISSUE #20: Possible functions.
How about some additional functions? Some are things I've used in the past, some are others I thought were sort of interesting, but might not be useful. Feel free to reject all of them.
### Eq stuff

``` purescript
equaling :: forall b. (Eq b) => (a -> b) -> a -> a -> Boolean
equaling = on eq
```

Useful for things like `intersectBy` or `nubBy`.

Could also use a newtype to change the value of equality.

``` purescript
newtype Diff a = Diff a

instance (Eq a) => Eq (Diff a) where
  (==) (Diff x) (Diff y) = x /= y
```
### Ord stuff

``` purescript
ascending :: forall a b. (Ord b) => (a -> b) -> a -> a -> Ordering
ascending = on compare

clamp :: forall a. (Ord a) => a -> a -> a -> a
clamp low high x = max low (min high x)

clamped :: forall a. (Ord a) => a -> a -> a -> Boolean
clamped low high x = low <= x && x <= high

descending :: forall a b. (Ord b) => (a -> b) -> a -> a -> Ordering
descending = on (flip compare)

max :: forall a. (Ord a) => a -> a -> a
max x y = if x <= y then y else x

min :: forall a. (Ord a) => a -> a -> a
min x y = if x <= y then x else y
```

`ascending` and `descending` are helpful for when you use something like `sortBy`. Also could use a newtype for changing the order of things, ala haskell:

``` purescript
newtype Desc a = Desc a

instance eqDesc :: (Eq a) => Eq (Desc a) where
  (==) (Desc x) (Desc y) = x == y

instance ordDesc :: (Ord a) => Ord (Desc a) where
  compare (Desc x) (Desc y) = compare y x
```
### Semiring stuff

``` purescript
inc :: forall a. (Semiring a) => a -> a
inc = add one
```
### Ring stuff

``` purescript
abs :: forall a. (Ord a, Ring a) => a -> a
abs x = if x <= zero then negate x else x

dec :: forall a. (Ring a) => a -> a
dec = flip sub one

sigNum :: forall a. (Ord a, ModuloSemiring a) => a -> a
sigNum x
  | x == zero = zero 
  | otherwise =  x `div` abs x
```

hdgarrood:
Some of this is already in purescript-posets, but it could potentially be more helpful to have it in Prelude.

### DIFF
--- a/docs/Prelude.md
+++ b/docs/Prelude.md
@@ -551,6 +551,14 @@ Instances must satisfy the following laws:
 
 `(*)` is an alias for `mul`.
 
+#### `succ`
+
+``` purescript
+succ :: forall a. (Semiring a) => a -> a
+```
+
+Adds `one` to a value.
+
 #### `Ring`
 
 ``` purescript
@@ -589,6 +597,23 @@ negate :: forall a. (Ring a) => a -> a
 
 `negate x` can be used as a shorthand for `zero - x`.
 
+#### `pred`
+
+``` purescript
+pred :: forall a. (Ring a) => a -> a
+```
+
+Subtracts `one` from a value.
+
+#### `abs`
+
+``` purescript
+abs :: forall a. (Ord a, Ring a) => a -> a
+```
+
+The absolute value of a value. Defined as
+`if x <= zero then negate x else x`.
+
 #### `ModuloSemiring`
 
 ``` purescript
@@ -619,6 +644,15 @@ laws:
 ```
 
 `(/)` is an alias for `div`.
+
+#### `sigNum`
+
+``` purescript
+sigNum :: forall a. (Ord a, Ring a, ModuloSemiring a) => a -> a
+```
+
+The 'sign' of a value. Defined as
+`if x == zero then zero else x / abs x`.
 
 #### `DivisionRing`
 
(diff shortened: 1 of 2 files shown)


garyb:
I wonder if we'd be better off with breaking some of these out into libraries like `purescript-ordering` for `Data.Ord`, etc instead. Adding these to the Prelude as they are are going to cause some problems: `min`, `max`, and `abs` are already defined in `purescript-math`, `succ` and `pred` will conflict with `purescript-enums`.

Also, this is just a personal opinion, but I think keeping the Prelude compact is something we should try to hold on to.

hdgarrood:
Ok, makes sense.

Exporting generalised versions of functions with the same name as `Math` functions doesn't both me very much though, particularly since I find I use integers far more often than floating point numbers, and nothing in Math works for `Int`. If we have generalised `min`, `max`, and `abs`, is there a reason to use the `Math` versions at all?

sharkdp:
I haven't looked at all the new functions in this PR, but I would strongly vote in favor of generalizing `max` and `min`! 

> Exporting generalised versions of functions with the same name as Math functions doesn't both me very much though, particularly since I find I use integers far more often than floating point numbers, and nothing in Math works for Int.

:+1:
It seems a shame that we have no `min/max :: Int -> Int -> Int` which we would get for free.

> If we have generalised min, max, and abs, is there a reason to use the Math versions at all?

Related: https://github.com/purescript/purescript-math/issues/8

hdgarrood:
Since I finally got around to publishing purescript-posets (now purescript-orders, given that we decided `Ord` should be a total ordering), I think we can close this.

PR #28, opened 2015-06-05
Go back to a simpler class for BooleanAlgebra

garyb:
As per discussion on IRC the other night, starting here: http://ircbrowse.net/browse/purescript?id=112058&timestamp=1433369439#t1433369439.

The flaw in this formulation we have now is `Bounded` has an `Ord` superclass constraint, which I'm pretty sure is wrong, if it's going to have one it should be no stronger than `Poset`. And unfortunately, if we don't introduce `Poset` in `Prelude` and just remove the `Ord` constraint, `Bounded` is completely lawless on its own.

### DIFF
--- a/docs/Prelude.md
+++ b/docs/Prelude.md
@@ -17,10 +17,6 @@ instance numUnit :: Num Unit
 instance eqUnit :: Eq Unit
 instance ordUnit :: Ord Unit
 instance boundedUnit :: Bounded Unit
-instance latticeUnit :: Lattice Unit
-instance boundedLatticeUnit :: BoundedLattice Unit
-instance complementedLatticeUnit :: ComplementedLattice Unit
-instance distributiveLatticeUnit :: DistributiveLattice Unit
 instance booleanAlgebraUnit :: BooleanAlgebra Unit
 instance showUnit :: Show Unit
 ```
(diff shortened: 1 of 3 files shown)


paf31:
:+1: Looks good. I don't think the `Ord` constraint is a big deal. We can still write some interesting instances, but we were going to move the poset stuff out into `-lattices` in any case.

garyb:
I was more concerned about `Ord` appearing in `Bounded` generally rather than within `BooleanAlgebra`'s superclasses - as it means we won't be able to use it for `BoundedLattice` now, for example.

docs/Prelude.md
```diff
@@ -809,18 +805,19 @@ Instances should satisfy the following law in addition to the `Ord` laws:
...
 
 ##### Instances
 ``` purescript
-instance latticeBoolean :: Lattice Boolean
-instance latticeUnit :: Lattice Unit
+instance booleanAlgebraBoolean :: BooleanAlgebra Boolean
+instance booleanAlgebraUnit :: BooleanAlgebra Unit
 ```
 
 The `Lattice` type class represents types that are partially ordered
```
hdgarrood: Should it be BooleanAlgebra here instead of Lattice?

garyb: Yep, thanks. Forgot to remake the docs.

PR #29, opened 2015-06-14
Ord docs

hdgarrood:
Update Ord and Ordering docs to specify that they describe total orderings now.

I am not sure why the documentation was previously being rendered below the instance list for type classes (this is why there are so many other changes in the rendered docs)... but I think the current behaviour of psc-docs, i.e. documentation then instance list, is nicer.


garyb:
:+1: 

I made a change to `psc-docs` recently to move the instance list down.

PR #31, opened 2015-07-12
Export unsafeCompare

andyarvanitis:
It would be nice to have `unsafeCompare` available to other packages, e.g. for modules providing other backend primitive types through the FFI. It would also be fine to move it to another package if it seems too dangerous to have in the Prelude.

`refEq` and `refIneq` would also be nice to export, but at least they're much simpler functions.

### DIFF
--- a/docs/Prelude.md
+++ b/docs/Prelude.md
@@ -975,4 +975,9 @@ instance showArray :: (Show a) => Show (Array a)
 instance showOrdering :: Show Ordering
 ```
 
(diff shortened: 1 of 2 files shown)


andyarvanitis:
Also, if this is ok, I'll update the doc.

andyarvanitis:
I went ahead and updated the doc, so this is ready to go if approved.

paf31:
Looks fine to me, thanks!

PR #36, opened 2015-08-12
Double backtick Remainder law so that nested backticks are code

mjgpy3:
Should resolve purescript#1368


hdgarrood:
Ah, great, I didn't know you could do this! I just tested and this works in Cheapskate too (which Pursuit uses):

```
λ: renderHtml $ toHtml $ C.markdown C.def "``a `mod` b == c``"
"<p><code>a `mod` b == c</code></p>\n"
```

paf31:
:+1: Thanks!

PR #37, opened 2015-08-13
Add Bounded Char instance

garyb:
This was previously an orphan in Data.Char, and also was wrong - the bottom and top values were reversed.

I'm not sure whether this is strictly speaking a breaking change or not, so not sure what we want to do about versioning here.

### DIFF
--- a/docs/Prelude.md
+++ b/docs/Prelude.md
@@ -836,6 +836,12 @@ _left-associative / precedence 4_
 
 Test whether one value is _non-strictly greater than_ another.
 
+#### `unsafeCompare`
+
+``` purescript
+unsafeCompare :: forall a. a -> a -> Ordering
+```
+
 #### `Bounded`
 
 ``` purescript
@@ -860,6 +866,7 @@ instance boundedBoolean :: Bounded Boolean
 instance boundedUnit :: Bounded Unit
 instance boundedOrdering :: Bounded Ordering
 instance boundedInt :: Bounded Int
+instance boundedChar :: Bounded Char
 instance boundedFn :: (Bounded b) => Bounded (a -> b)
 ```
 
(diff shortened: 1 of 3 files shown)


garyb:
I implemented the values through the FFI to avoid having to move `fromCharCode` over also, by the way.

paf31:
Sorry, Just woke up, didn't mean to merge this. Brain is on auto-pilot. I've reverted it for now. PR looks good though.

garyb:
Hah, no problem. Merging is fine anyway, the question is just about whether we want to bump to v0.2.0 for this... might be a bit painful as I've still not got a working auto-version-bump tool.

paf31:
I think we have to ... right? If someone already has Prelude and installs the latest `strings`, their code will be busted :(

garyb:
Removing the `Bounded Char` instance from `strings` was a breaking change though, I versioned that one as such... it's more a case of it might break for people with the older `strings` if they get the newer `prelude`.

garyb:
If the previous `Bounded Char` instance wasn't entirely broken (inverted `top` and `bottom`) I'd probably agree it is a breaking change, but I have to assume nobody was using it previously...

garyb:
And one final point - the old `strings` does compile with the new `Prelude`, it's only a problem when the instance search is triggered.

paf31:
Ah ok. Maybe we can get away with it then? I don't think any code uses this instance yet.

But then, we've been bitten by things like this before :(

paf31:
Sorry, now I see. Yes, minor version bump for Prelude, major for `strings` makes sense.

paf31:
I'll revert my revert then?

garyb:
Haha, sure :)

paf31:
Alright done, I'll tag.

paf31:
So we need to bump everything which depends on `strings`, right? How many things is that?

garyb:
It's not too bad, I've already started doing it:

```
purescript-enums
purescript-quickcheck
purescript-maps
purescript-datetime
purescript-sets
purescript-graphs
```

garyb:
(enums needed the prelude update first)

PR #41, opened 2015-09-12
Modularised Prelude

garyb:
This is not for merging just yet, as this formulation has some changes that will have impact elsewhere:

The following modules are currently defined in `purescript-control` but sans the actual classes. I've merged them in here, so I guess if we did this we'd want to remove them from `control`? Either that or reverse the dependency and `Prelude` would become something more like an even more minimal `base`.
- `Data.Functor`
- `Control.Apply`
- `Control.Applicative`
- `Control.Bind`
- `Control.Monad`

It seems like `Data.Functor` should belong in `-functors` now we have it, but that's not a great option without some work and dependency shuffling - it depends on `tuples` and `foldable-traversable`.

Other breaking changes we talked about:
- Removed the `return` alias for `pure`
- Removed the `(++)` alias for `append`

Some we didn't discuss, but seem worth considering:
- Renamed `liftM1` to just `liftM`. I guess if we were going to include higher arity versions then keeping the 1 would make sense alternatively.
- Renamed the `lift2`...`lift5` functions to `liftAn` to make them less ambiguous and to match `liftA1`.
- Moved `when` and `unless` into `Control.Applicative` and relaxed the `Monad` constraint to `Applicative`.

Some non-breaking changes/enhancements:
- Each `Control` module exports its `Control` dependencies, so if you import `Applicative` then you get `Functor` and associated operators in scope too.
- Same for the number hierarchy
- Added instances for functions in the style of `(X b) => X (a -> b)` as we had them on some classes and not others currently

There's a lot of stuff here, maybe we don't want any of it - that's fine, it was mainly to pass some time while on the road :smile:


src/Prelude.purs
```diff
@@ -1,108 +1,29 @@
 module Prelude
-  ( Unit(), unit
-  , ($), (#)
-  , flip
-  , const
-  , asTypeOf
+  ( module P
```
paf31, line 2: Whoa, this works now? :smile:

garyb, line 2: Yep! You need to `import qualified _ as _` though, `import _ as _` doesn't work which perhaps should be added too.

paf31:
:+1: Looks nice.

garyb:
I think I'll come back to this later. Perhaps we can squeeze in operator aliases just before we wrap up 0.8 - it'd be nice to not to have to break Prelude all over again when operator aliases land - probably can start tagging libraries as `v1.0.0` once we get those in too?

PR #43, opened 2015-10-20
Add Natural type synonym

### LINKED ISSUE #33: Add `Natural` type synonym?
For natural transformations:

``` purescript
type Natural f g = forall a. f a -> g a
```

Currently we have this in `Data.Coyoneda` but it's much more widely applicable and pretty generic, yet I'm not sure it's worth a module of its own as there aren't really any operations that go with it (I think?).

honza:
Closes #33

### DIFF
--- a/docs/Prelude.md
+++ b/docs/Prelude.md
@@ -983,4 +983,20 @@ instance showArray :: (Show a) => Show (Array a)
 instance showOrdering :: Show Ordering
 ```
 
+#### `NaturalTransformation`
+
+``` purescript
+type NaturalTransformation f g = forall a. f a -> g a
+```
+
+A natural transformation
+
+#### `Natural`
+
+``` purescript
+type Natural f g = NaturalTransformation f g
+```
+
+Alias for `NaturalTransformation`
+
 
--- a/src/Prelude.purs
+++ b/src/Prelude.purs
@@ -25,6 +25,8 @@ module Prelude
   , BoundedOrd
   , BooleanAlgebra, conj, disj, not, (&&), (||)
   , Show, show
+  , NaturalTransformation(..)
+  , Natural(..)
   ) where
 
 -- | The `Unit` type has a single inhabitant, called `unit`. It represents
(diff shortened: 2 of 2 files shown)


docs/Prelude.md
```diff
@@ -983,4 +983,20 @@ instance showArray :: (Show a) => Show (Array a)
...
+``` purescript
+type NaturalTransformation f g = forall a. f a -> g a
+```
+
+A natural transformation
+
+#### `Natural`
+
+``` purescript
+type Natural f g = NaturalTransformation f g
```
paf31, line 997: I think it'd be nice to pick one name and stick with it. I'd pick `NaturalTransformation` since the extra verbosity doesn't cost much and it's explicit.

paf31:
:+1: Thanks!

docs/Prelude.md
garyb, line 997: Hah I argued for the exact opposite in https://github.com/purescript/purescript-free/pull/37 :smile:

One minor problem, if we don't bump the version as a breaking change bump after merging this it will probably break a bunch of stuff. Halogen for sure.

honza: I just grabbed the code from [purescript-free](https://github.com/purescript/purescript-free/blob/master/src/Data/NaturalTransformation.purs). Let me kill the shorter version.

paf31, line 997: My other complaint about `Natural` is that we might want to reserve that name in something like `Prelude` for a natural number type.

Also, let me play devil's advocate for a second and say - since we can define a synonym (or not!) anywhere, what's the advantage of adding this to Prelude?

ghost, line 997: For a shorter name I think `Transform` might be good. I agree that it would be nice to have `Natural` available for some numerics thing at some point. I like `Transform` for two reasons: 1) the "natural" part is kind of superfluous in my opinion since nobody talks about unnatural transformations and 2) there's a precedent for something similar sounding with [(n,k)-transfors](http://ncatlab.org/nlab/show/transfor#special_cases), which are sometimes used to describe the sequence of functors, natural transformations, …. Homotopy would be another nice short name but probably harder to justify and maybe a bit obscure in a programming context.

honza, line 997: One advantage of adding this to the Prelude is to increase the Prelude's didactic function. When starting out with Purescript, I found it exceedingly instructive to simply sit down and read through the Prelude.

paf31:
I think we have `Natural`/`~>` now, don't we?

hdgarrood:
I don't see it in here: https://github.com/purescript/purescript-prelude/blob/e27709762b30334c09a76e255e525890d72c65c1/src/Prelude.purs

garyb:
We haven't defined the operator anywhere yet actually. Adding it now with the `type (~>)` sounds great to me though.

honza:
Closing in favor of 20004f9

PR #48, opened 2015-12-19
0.8 updates

garyb:


garyb:
Inlining will need updating in the compiler too actually. We can probably support both old and new style Preludes though I think.

hdgarrood:
I'm not a huge fan of `esopmoc` for `>>>`, and `ylppa` for `#`, even if they will rarely ever get used. What about `composeBack` or `composeFlipped`?

src/Prelude.purs
```diff
@@ -128,16 +94,13 @@ class Semigroupoid a where
...
-
--- | `(<<<)` is an alias for `compose`.
-(<<<) :: forall a b c d. (Semigroupoid a) => a c d -> a b c -> a b d
-(<<<) = compose
+infixr 9 compose as >>>
 
 -- | Forwards composition, or `(<<<)` with its arguments reversed.
-(>>>) :: forall a b c d. (Semigroupoid a) => a b c -> a c d -> a b d
-(>>>) = flip compose
+esopmoc :: forall a b c d. Semigroupoid a => a b c -> a c d -> a b d
```
paf31: :scream: `composeBackwards` perhaps?

garyb:
Yeah, I figured maybe it was a bit cute ;) was thinking of `cons`/`snoc`, but `xFlipped` sounds fine to me.

src/Prelude.purs
```diff
@@ -167,6 +130,18 @@ instance categoryFn :: Category (->) where
...
   map :: forall a b. (a -> b) -> f a -> f b
 
+infixl 4 map as <$>
+
+-- | `pam` is `map` with its arguments reversed. For example:
+-- |
+-- | ```purescript
+-- | [1, 2, 3] <#> \n -> n * n
+-- | ```
+pam :: forall f a b. Functor f => f a -> (a -> b) -> f b
```
paf31: `mapBackwards`? This one is easier to remember, but I don't think we lose much by spelling it out.

garyb:
I think `backwards` is problematic, as `composeBackwards` would actually be forwards composition, so I suggest `flipped` throughout for consistency, as that's the definitions of them all too.

src/Prelude/Internal.purs
```diff
@@ -0,0 +1,39 @@
...
+-- | length $ groupBy productCategory $ filter isInStock $ products
+-- | ```
+-- |
+-- | is equivalent to:
+-- |
+-- | ```purescript
+-- | length (groupBy productCategory (filter isInStock products))
+-- | ```
+-- |
+-- | `($)` is different from [`(#)`](#-2) because it is right-infix instead of
```
paf31: What's the `-2` here?

src/Prelude.purs
hdgarrood: I'm not sure the "backwards" suffix works so well here. `compose` sort of implies an order, whereas I don't think `map` does.

src/Prelude/Internal.purs
garyb: No idea, it was copy/pasted. I should probably update the comment to reflect the named functions anyway?

paf31:
What about something like `mapped` or `mapping`?

garyb:
Updated to remove oddly named functions in favour of `flipped` throughout. `mapped` or something might be better in that case, but I figure really the flipped versions exist for operator convenience anyway, so the naming isn't super important.

I also added `(=<<)` to `Prelude`: I almost always use that form of it rather than `(>>=)`, as it plays better with `(<<<)`, and since we have other flipped functions, why not?

Will squash the commits before merging if you're happy. Figured it's easier to review the incremental diff.

src/Prelude.purs
```diff
@@ -314,9 +270,16 @@ liftA1 f a = pure f <*> a
 -- |    y <- m2 x
 -- |    m3 x y
 -- | ```
-class (Apply m) <= Bind m where
+class Apply m <= Bind m where
   bind :: forall a b. m a -> (a -> m b) -> m b
 
+infixl 1 bind as >>=
+
+bindFlipped :: forall m a b. Bind m => (a -> m b) -> m a -> m b
```
paf31, line 283: How about 

``````
-- | `bindFlipped` is `bind` with its arguments reversed. For example:
-- |
-- | ```purescript
-- | print =<< random
-- | ```
``````

?

Does the fixity here allow you to do things like `f <<< g <<= xs`?

garyb, line 283: Oops, missed that comment, I'll add it.

Yep, although with `bind` rather than `extend` :wink:

paf31:
Looks good to me (I added one more comment).

What's the plan for releasing these? RCs?

src/Prelude.purs
paf31, line 283: D'oh :smile:

garyb:
I think we can just go ahead and release them, and just make it clear in the release they're for the RC compiler. We're not going to break anything we have here, and I'll be explicit about every import/export in case we do remove/make `(..)` warn.

paf31:
Yes, the only issue is if someone does `bower i purescript-prelude` in an empty project.

paf31:
But I'm fine with that, we probably just need to document it in the release notes and/or readme.

garyb:
Well, we could hold off releasing any of them until 0.8-proper is released? Releasing everything as `-rc` first and then updating it all again afterwards wasn't much fun last time :wink:

garyb:
Updated again with the comment for `bindFlipped`. I also made the `Ordering(..)` export explicit.

hdgarrood:
Yeah, I think if re-releasing RCs is too much effort, then using either `bower link` or explicitly using `master` in `bower.json` are better options than releasing proper versions that don't work with any (non-RC) compiler yet.

garyb:
`master` is definitely not an option, `bower` caches the first thing it downloads from a tag/branch, so it causes all kinds of mayhem over time.

garyb:
I'm happy enough to just `bower link` and wait until the compiler is released though; it's not imperative that the libraries get updated along with the compiler this time, as nothing is broken... there will just be a few warnings to ignore :wink:

hdgarrood:
Ah, I didn't know that. That's disappointing. So I guess you might as well just use a commit SHA then, rather than `master`?

hdgarrood:
In general, I mean.

garyb:
Yeah. It's the same deal with `*` - it's really odd actually, it seems like a bug even, as it's definitely not what anyone would actually want these values to do.

PR #55, opened 2016-01-18
fixes #53: properly escape output of Show instance for String

### LINKED ISSUE #53: Show instance for String produces invalid escape sequences
Since it's implemented via `JSON.stringify`, JavaScript escape sequences are generated. As an example, `show "\0"` generates `"\u0000"`, which is not a valid PureScript string.

michaelficarra:
Fixes #53.


paf31:
:+1: Seems right to me.

PR #56, opened 2016-01-19
fix .jshintrc and Travis CI

michaelficarra:
The `globalStrict` option appears to have been deprecated in favour of `strict: "true"` according to [this comment in jshint](https://github.com/jshint/jshint/blob/2e0026f8f461425bb05da966570708002a9f4f0a/src/options.js#L361-L370). This has been the cause of recent Travis CI failures.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -5,7 +5,7 @@
   "freeze": true,
   "funcscope": true,
   "futurehostile": true,
-  "globalstrict": true,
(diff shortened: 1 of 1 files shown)


garyb:
Thanks! I've encountered this on other repos that have jshint setup too so was going to mention on your other PRs, but just didn't get around to it yesterday.

PR #58, opened 2016-01-20
define Unit and unit using foreign imports

michaelficarra:
As suggested by @paf31 in https://github.com/purescript/purescript-prelude/pull/57#issuecomment-172937032. Closes #57.

### DIFF
--- a/src/Prelude.js
+++ b/src/Prelude.js
@@ -3,6 +3,8 @@
 
 // module Prelude
 
+exports.unit = void 0;
+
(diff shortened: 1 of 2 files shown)


paf31:
:+1: Looks good to me.

garyb:
I'm not sure whether this is a breaking change or not - it definitely shouldn't be, but it seems a little dodgy to switch it out in a minor version bump, so I'll incorporate this into the 0.8 stuff.

PR #59, opened 2016-01-20
add missing `$` in docs for ($); mirrors docs for (#)

michaelficarra:


paf31:
`pulp docs` please? Looks good.

michaelficarra:
@paf31 How's this? I've split it up into 3 commits because the docs weren't previously being built with a recent version of PureScript.

hdgarrood:
I think that new $ is unnecessary though; as far as I can tell, none of the # are. I think I'd leave it how it is. If I was writing code under normal circumstances, I would prefer the version without the extra $.

michaelficarra:
@hdgarrood I still think it's appropriate because the first equivalence (`a $ b $ c $ d $ x = a $ (b $ (c $ (d $ x)))`) simply shows the associativity and the second equivalence (`a $ (b $ (c $ (d $ x))) = a (b (c (d x)))`) shows how `$` is function application.

LiamGoodacre:
This should be closed now: the documentation for `$` doesn't appear list the bit that was changed anyway.

Issue #62, opened 2016-03-13
Break up BooleanAlgebra


paf31:
cc @freebroccolo

We discussed modifying the `BooleanAlgebra` hierarchy on IRC.

I'd like to support some simple overloading of the boolean operations, while not splitting things into a full hierarchy of `Lattice`s, which I think would be fine in another library.

I propose:
- Split out `HeytingAlgebra`
- Make `Bounded` a subclass of `Ord`
- Copy the `Bounded` members into `HeytingAlgebra`

as follows:

``` purescript
-- See https://en.wikipedia.org/wiki/Heyting_algebra#Bounded_lattice_with_an_implication_operation
--
-- a ==> a = tr
-- a && (a ==> b) = a ==> b
-- b && (a ==> b) = b
-- a ==> (b && c) = (a ==>b) && (a ==> c)
--
-- plus the laws for a bounded lattice
--
-- plus `not a = a ==> fa`
class HeytingAlgebra h where
  tr :: h
  fa :: h
  implies :: h -> h -> h
  conj :: h -> h -> h
  disj :: h -> h -> h
  not :: h -> h

infix ==> as implies -- for the purposes of writing out the laws above

infix && as conj
infix || as disj

-- adds the law of the excluded middle
class (HeytingAlgebra b) <= BooleanAlgebra b
```

This allows us to interpret the operations of Boolean algebra in a variety of settings:
- `Boolean` itself
- Tuples
- Power sets of `Finite` types
- The interval `[0, 1]` with `min` and `max` operations (Heyting but not Boolean)
- Three-valued logic (Heyting but not Boolean)
- In general, we can create a `HeytingAlgebra` from any `Bounded` type with an order-reversing involution.

etc.

The tricky bit is that there is obviously an interaction with the `Ord` hierarchy here, but I'm trying to hide it in order to get something useable without refining the hierarchy too much, hence the formulation of `HeytingAlgebra` in terms of implication and conjugation, instead of the underlying order.

garyb:
I'm not crazy about the names `tr` and `fa`, `fa` in particular is almost certainly going to be shadowed... but then I don't have any suggestions of alternatives either really.

Is there a counterexample of why we shouldn't use `Ord` or `Bounded` as a superclass here? We could use `top` and `bottom` for `tr` and `fa` perhaps then.

paf31:
> Is there a counterexample of why we shouldn't use Ord or Bounded as a superclass here? 

[...]

PR #64, opened 2016-03-18
Class updates

### LINKED ISSUE #61: ModuloSemiring laws
I'm opening this ticket following a discussion with @paf31 on IRC. Here's what we've got:

``` purescript
-- | The `ModuloSemiring` class is for types that support addition,
-- | multiplication, division, and modulo (division remainder) operations.
-- |
-- | Instances must satisfy the following law in addition to the `Semiring`
-- | laws:
-- |
-- | - Remainder: ``a / b * b + (a `mod` b) = a``
class (Semiring a) <= ModuloSemiring a where
  div :: a -> a -> a
  mod :: a -> a -> a
```

This "law", however, is not valid unless we intend for the integers to fail to model the `ModuloSemiring` structure. In fact, the only instance of `ModuloSemiring` that actually follows this law presently is `Unit`. The problem is that nowhere has the behavior of division by zero been accounted for....
#### counterexample

Here's an example of where the law fails for `Number`. Fix `a : Number`, and let `b` be `0`. Then, we have the following chain of equations:
1. `(a / 0) * 0 + mod a 0`
2. `==> Infinity * 0 + mod a 0`
3. `==> 0 + 0`
4. `==> 0`

Therefore, `a` must be `0` for any `a`, which is clearly false.
#### discussion

what should we do? Unless I've made a mistake, this law is definitely wrong / not the intended definition of a `ModuloSemiring`. Can we fix it though? One option would be to say that it must only hold for non-zero `b`, but is that something we want to say in the type class itself? I'm curious what people think.

@garyb @paf31

garyb:
Resolves #61, #62

I went with the name `EuclideanRing` rather than `EuclideanDomain`, it seemed to fit better with our other `Ring`-terminology classes, and seems to be an acceptable name for the structure also.

Re: `HeytingAlgebra`. I did go with `largest` and `smallest` in the end. Although that strays from truth terminology it shouldn't really matter, as it's a lattice too, and from that point of view those names are just as reasonable.

I also added the Kleene 3VL for `Ordering`... it's true that `(&&)` and `(||)` don't make a whole lot of sense from an ordering point of view, but it certainly satisfies the necessary laws. If it still seems like a bad idea that's fine, I can take it out again.


paf31:
:+1: Thanks for working on these. Looks great to me, for the most part:

I would keep the standard names, honestly, and I'd probably pick something like `truth` and `falsehood` for `HeytingAlgebra`, or `tt` and `ff` if we wanted terse names. As for the three-valued logic, I think it would be better to leave that for a library. Is there a reason to include it in Prelude? It doesn't really belong with `Ordering`, and the only connection seems to be that both types have three values.

garyb:
I wasn't suggesting we use `Ordering` as a 3VL, only that by treating it as one we get a fully law abiding instance. The motivation is `not` again, as especially with a `(HeytingAlgebra b) => HeytingAlgebra (a -> b)` instance there are some nice things that can be done. I guess my point is: there is a lawful instance, it's sometimes useful, so why not?

I'm not sure I agree that `true` and `false` are necessarily better terms that `top` and `bottom` or `largest` and `smallest`... given that it's a bounded lattice with implication operation here, which doesn't seem especially truthy in particular. But anyway I've realised, it doesn't matter! My concern was about cluttering up the namespace of anyone who imports the `Prelude` (given that one and two letter names are the common for local variables), but since the `Prelude` is re-exporting now, I can just not import `tt` / `ff` and that can be avoided.

paf31:
Given the existence of `Down` to turn an ordering around, I'd prefer not to add a potentially confusing `HeytingAlgebra Ordering` instance to `Prelude`. What's a case where you'd use `not` with the function instances? I suspect it'd be just as simple to use `Down` in those cases.

paf31:
Also, why not reuse the names `top` and `bottom`, since they're in different modules anyway?

garyb:
Yeah, I guess, or not `Down` but `invert`:

```
xs' = sortBy (not compare) xs
xs' = sortBy (\x -> invert <<< compare x) xs
```

The first is still much nicer though :wink:

I think I'd rather go with `tt` / `ff` than introduce an alternative `top` / `bottom`, as then the question will be why it isn't just `Bounded` in the first place.

edit: `invert` not `reverse`

paf31:
It's shorter, but I don't know if it's nicer. I would find it quite hard to figure out exactly which `HeytingAlgebra` the `not` referred to.

I don't think the existence of an involution which happens to be `not` for some unrelated `HeytingAlgebra` is enough to justify the instance.

> I think I'd rather go with tt / ff than introduce an alternative top / bottom, as then the question will be why it isn't just Bounded in the first place.

Well, at least we know the answer :smile: I'm not fussed either way though.

garyb:
Ok, it's gone.

How does the rest of it look, in particular I wanted to check that:
- `degree a = |a|` for `Int`
- `implies = eq` for `Boolean`

Are good implementations?

paf31:
`p ==> q = not p || q`, so

``` purescript
implies true true = true
implies false true = true
implies true false = false
implies false false = true
```

The `degree` implementation is good, see https://en.wikipedia.org/wiki/Euclidean_domain#Examples.

garyb:
Oops, that's actually what I used for the 3VL too as I found it elsewhere... thanks.

paf31:
This looks good to me, but I'd be very grateful if @jonsterling and/or @freebroccolo would take a look the at the new hierarchy before it gets merged.

garyb:
Oh, too late :smile: but we can always change stuff again while it's still in RC mode.

paf31:
Oh haha, I should have looked first. That's fine, as you say, we're still only at RC 2.

PR #67, opened 2016-03-30
fixes #65: show implementation for strings/chars containing U+0007

### LINKED ISSUE #65: Bad escape in Show
Sorry if this is the wrong place - but I think it's maybe multiple issues.

I get the following error from `psc-bundle` 0.8.3 (only on Windows!):

```
Error in module Prelude (Foreign):

  The module could not be parsed:
  lexical error @ line 216 and column 14
```

This points at: https://github.com/purescript/purescript-prelude/blob/4c97cecbf041f27b83f7ae9d39fc2d9b24a097f4/src/Prelude.js#L216

The same line is present in the 1.0-rc modularised version. As far as I can see (eg [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Grammar_and_types#Using_special_characters_in_strings)) `\a` isn't actually a JavaScript string escape sequence, which is why I think it's an issue here - can we remove it? On the other hand the behaviour for an unknown escape should be to pass it through (`\a` means precisely `a`), which would be an issue for language-javascript, probably post https://github.com/purescript/purescript/issues/1966.

michaelficarra:
`"\a"` in JavaScript is `"a"`, not `"\x07"` like in PureScript/Haskell. Fixes #65.

### DIFF
--- a/src/Data/Show.js
+++ b/src/Data/Show.js
@@ -15,7 +15,7 @@ exports.showCharImpl = function (c) {
   var code = c.charCodeAt(0);
   if (code < 0x20 || code === 0x7F) {
     switch (c) {
-      case "\a": return "'\\a'";
+      case "\x07": return "'\\a'";
       case "\b": return "'\\b'";
       case "\f": return "'\\f'";
       case "\n": return "'\\n'";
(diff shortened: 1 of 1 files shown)


garyb:
Thanks! I'll make a patch for the non-RC version from this too.

PR #70, opened 2016-04-27
Add `return` as a synonym for `pure`

tslawler:
This seems to have gotten dropped between 0.1.5 and 1.0.0-rc.1 -- not sure if this was intentional, but it seems like it ought to remain in, if only to make an upgrade less of a breaking change for people using `return` in do-notation.

### DIFF
--- a/src/Control/Applicative.purs
+++ b/src/Control/Applicative.purs
@@ -1,5 +1,5 @@
 module Control.Applicative
-  ( class Applicative, pure
+  ( class Applicative, pure, return
   , liftA1
   , unless, when
   , module Control.Apply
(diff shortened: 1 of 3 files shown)


garyb:
It is intentional, the same with `(++)` being dropped so we only have `(<>)` - having multiple names for the same thing is something we want to avoid in the core libraries.

Since the 1.0.0 libraries are going to be released with psc 0.9 it's almost certainly going to involve breaking changes for everyone, so we may as well get as much pain out of the way all at once!

tslawler:
In that case are you also going to remove `otherwise` as a synonym for `true`? ;)

garyb:
Well, you got me there. :laughing: 

I guess that's different because it's a syntactic nicety, and it's probably the only one we'll have. `return` is no better than `pure` in that regard, since it doesn't mean "return" in the sense that any non-Haskell programmer would be familiar with.

hdgarrood:
Shall we close this?

PR #72, opened 2016-05-13
Fix Show Number for integers outside the Int range

### LINKED ISSUE #71: Show Number can produce invalid Number literals
For integers outside the `Int` range, you might get an `Int` literal instead of a `Number` (note that `Number` literals are required to either contain a decimal point or an exponent).

For example:

```
> show (1e10 :: Number)
10000000000
> :t 10000000000
[Integer value is out of range...]
```

I suggest that instead of the bitwise-or-with 0 check, we check that the produced string either contains a decimal point or an 'e' or 'E', and if not, append ".0".

hdgarrood:
Fixes #71

### DIFF
--- a/src/Data/Show.js
+++ b/src/Data/Show.js
@@ -7,8 +7,10 @@ exports.showIntImpl = function (n) {
 };
 
 exports.showNumberImpl = function (n) {
-  /* jshint bitwise: false */
-  return n === (n | 0) ? n + ".0" : n.toString();
+  var str = n.toString();
+  return str.indexOf("e") === -1 && str.indexOf(".") === -1 ?
+            str + ".0" :
+            str;
 };
 
 exports.showCharImpl = function (c) {


garyb:
:+1:

paf31:
👍

michaelficarra:
Other possible outputs from `ToString` are `"NaN"` and `"Infinity"`. See https://tc39.github.io/ecma262/#sec-tostring-applied-to-the-number-type.

garyb:
Oh damn, good point. We don't have those as literals either, which makes `Show` a bit awkward if we're following the way it's supposed to work.

hdgarrood:
Oops, thanks for the reminder! I think it's probably fine to just print NaN or Infinity in those cases. The informal Show law isn't adhered to particularly strongly anyway, and I can't remember the last time I copied the output of show into a source file either. As far as I can tell, the alternatives would just be something like "1/0" or "0%0" which is probably going to be a bit confusing.

garyb:
Yeah, that makes sense. We don't have a `Read`, so it's less important in that regard too.

michaelficarra:
@hdgarrood But we won't print `NaN` or `Infinity` in those cases. After this PR, we'll print `NaN.0` and `Infinity.0` :sweat_smile:.

texastoland:
This one works with fewer corner cases:

``` js
exports.showNumberImpl = function (n) {
  var  str = n.toString();
  return isNaN(str + ".0") ?  str :  str + ".0";
}
```

`isNaN()` implicitly calls `Number()` which returns `NaN` if it can't parse.

hdgarrood:
Yeah, I did realise that, I just haven't gotten around to sending another one yet ;)

So if we take what we have now but check for and handle NaN and Infinity explicitly, does that sound sensible?

hdgarrood:
Haskell prints Infinity and NaN too, incidentally, even though these don't work as Haskell code.

michaelficarra:
This is how I would write it:

``` js
exports.showNumberImpl = function (n) {
  return isFinite(n) && Math.floor(n) === n ? n + ".0" : "" + n;
};
```

**edit:** This doesn't work for very large integers that are printed using scientific notation.

hdgarrood:
@michaelficarra That won't work for integral values; we always want either a ".0" on the end, or a "e" somewhere inside the value, as if a literal can be parsed as an integer, it will be an `Int`, not a `Number`.

michaelficarra:
@hdgarrood I just had the condition reversed. It's fixed now.

AppShipIt:
@michaelficarra `1e30` doesn't work?

PR #73, opened 2016-05-14
Remove redundant imports in Data.Ord

lexi-lambda:
Exactly what it says on the tin: removes two unused imports in `Data.Ord` that `psc` was complaining about.


paf31:
👍 LGTM, thanks!

garyb:
Thanks!

PR #76, opened 2016-05-17
Fix show number again

hdgarrood:
cc @AppShipIt @michaelficarra, refs #72

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ install:
   - npm install -g bower
   - npm install
 script:
-  - npm run build
+  - npm run build && npm test
 after_success:
 - >-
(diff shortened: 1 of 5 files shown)


michaelficarra:
@hdgarrood See my comment in #72. I'd prefer that implementation.

hdgarrood:
Unfortunately, your implementation doesn't work for numbers with exponents:

```
Error: For 1e+21, expected 1e+21, got: 1e+21.0.
```

texastoland:
Agreed it's nice (less coercion) but that's the case I found too.

michaelficarra:
Ah, got it. :+1: then.

garyb:
Thanks everyone involved in this!

PR #77, opened 2016-05-18
Remove unused imports. Fixes UnusedImport warnings.

jonboiser:
I've been noticing a lot of these warnings when compiling projects lately. Just wanted to fix a few.


garyb:
There was another PR open (#73) for this already, so thanks for the contribution but I merged that one instead.

PR #78, opened 2016-05-21
Add abs, signum. resolves #66

### LINKED ISSUE #66: A proposal for generalized `abs`
I think it would be nice to get some concept of `abs` which works for not only `Int` and `Number`, but also more exotic numeric types, like `Rational`, `Complex`, and even `Quaternion` (when we eventually implement it ;).

Given that we (well, I) want this to work for numeric types which don't have total orders (such as `Complex`), I thought a good starting point would be to take laws from https://en.wikipedia.org/wiki/Absolute_value but remove all those to do with ordering, so:

``` purescript
class Semiring a <= Absolute a where
  abs :: a -> a
```

Laws:
- Positive-definiteness: `abs x == zero iff x == zero`
- Multiplicativeness: `abs (x * y) == abs x * abs y`
- Idempotence: `abs (abs x) == abs x`

We could add more laws by making the constraint a bit stronger, to `Ring` if we want those that mention subtraction, or maybe even `EuclideanRing` if we want those that mention division. But the above seems a good starting point, I think.

hdgarrood:
See discussion in #66.

### DIFF
--- a/src/Data/Ord.purs
+++ b/src/Data/Ord.purs
@@ -8,6 +8,8 @@ module Data.Ord
   , min, max
   , clamp
   , between
+  , abs
+  , signum
   , module Data.Ordering
   ) where
 
@@ -15,7 +17,7 @@ import Data.Eq (class Eq)
 import Data.Function (on)
 import Data.Ord.Unsafe (unsafeCompare)
 import Data.Ordering (Ordering(..))
-import Data.Ring (negate)
+import Data.Ring (class Ring, zero, one, negate)
 import Data.Unit (Unit)
 import Data.Void (Void)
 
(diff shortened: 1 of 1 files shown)


src/Data/Ord.purs
```diff
@@ -149,3 +151,13 @@ between low hi x
   | x < low = false
   | x > hi = false
   | true = true
+
+-- | The absolute value function. `abs x` is defined as `if x > zero then x
+-- | else negate x`.
+abs :: forall a. (Ord a, Ring a) => a -> a
+abs x = if x > zero then x else negate x
```
garyb: Should we use `x >= zero` here so we don't end up with signed zero weirdness? It would probably make sense for `abs 0` to return `0` rather than `-0`!

`0 === -0` is true, but `1/0 === Infinity`, `1/-0 === -Infinity` and `Infinity !== -Infinity` so it's potentially observable other than that too.

hdgarrood: I was about to say, isn't `negate x` just `zero - x`, but of course it isn't for `Int` and `Number` is it because of inlining! Yes, definitely.

PR #79, opened 2016-05-22
Add natural transformation synonym & operator

### LINKED ISSUE #23: Testing
This applies to all the core libraries, not just prelude, but this the first one in the hierarchy where the problem of circular `devDependencies` occurs.

I've been thinking about ways we might resolve this, one approach we had previously was to just remove the offending duplicate from `bower_components` when building, and actually, something along these lines is by far my preference, although this time using a `postinstall` script in `.bowerrc` so it happens automatically as part of the install process.

It's so much simpler than any other alternative, with basically no downsides I can see (other than being aesthetically offensive :wink:) - `bower` certainly doesn't care if you have circular dependencies.

If you `bower install x` something where `x` has `devDependencies` the `devDependencies` are not installed. `bower install` in the current project, without the `--production` flag is the only time `devDependencies` are brought in, and in this situation `.bowerrc` will be picked up for the project, so we can definitely do the removal there.

All the other ways I've looked at, having a separate `bower.json` in `tests/`, etc. make it really awkward to run tests as part of the normal `gulp` or `grunt` build (actually, impossible in the `grunt` case, as you have to run `grunt` in the directory the `Gruntfile.js` lives) and then you end up having to maintain two sets of bower dependencies too.

garyb:
Resolves #23 

Once this is merged `purescript-free` will need an update, as currently that library provides a module `Data.NaturalTransformation`.

### DIFF
--- a/src/Data/NaturalTransformation.purs
+++ b/src/Data/NaturalTransformation.purs
@@ -1,0 +1,19 @@
+module Data.NaturalTransformation where
+
+-- | A type for natural transformations.
+-- |
+-- | A natural transformation is a mapping between type constructors of kind
+-- | `* -> *` where the mapping operation has no ability to manipulate the
+-- | inner values.
+-- |
+-- | An example of this is the `fromFoldable` function provided in
+-- | `purescript-lists`, where some foldable structure containing values of
+-- | type `a` is converted into a `List a`.
+-- |
+-- | The definition of a natural transformation in category theory states that
+-- | `f` and `g` should be functors, but the `Functor` constraint is not
+-- | enforced here; that the types are of kind `* -> *` is enough for our
+-- | purposes.
+type NaturalTransformation f g = forall a. f a -> g a
+
+infixr 4 type NaturalTransformation as ~>
--- a/src/Prelude.purs
+++ b/src/Prelude.purs
@@ -15,6 +15,7 @@ module Prelude
   , module Data.Function
   , module Data.Functor
   , module Data.HeytingAlgebra
+  , module Data.NaturalTransformation
   , module Data.Ord
   , module Data.Ordering
   , module Data.Ring
@@ -42,6 +43,7 @@ import Data.Field (class Field)
 import Data.Function (const, flip, ($), (#))
 import Data.Functor (class Functor, map, void, ($>), (<#>), (<$), (<$>))
 import Data.HeytingAlgebra (class HeytingAlgebra, conj, disj, not, (&&), (||))
+import Data.NaturalTransformation (type (~>))
 import Data.Ord (class Ord, compare, (<), (<=), (>), (>=), comparing, min, max, clamp, between)
 import Data.Ordering (Ordering(..))
 import Data.Ring (class Ring, negate, sub, (-))


src/Data/NaturalTransformation.purs
```diff
@@ -0,0 +1,14 @@
...
+-- |
+-- | A natural transformation allows a functor to be transformed into another
+-- | without changing the internal structure. This is a little similar to the
+-- | `map` operation of a `Functor`, although "one level up". In the case of
+-- | `map`, the functor is preserved while the inner values are re-mapped,
+-- | whereas here the inner values are preserved and the functor is changed
+-- | instead.
+type NaturalTransformation f g = forall a. f a -> g a
+
+infixr 4 type NaturalTransformation as ~>
```
garyb, line 19: I figured this gives us enough room to slide `->` underneath while still not making it the lowest precedence, when we decide what that should be.

src/Data/NaturalTransformation.purs
```diff
@@ -0,0 +1,14 @@
+module Data.NaturalTransformation where
+
+-- | A type for natural transformations between functors (although the `Functor`
```
paf31: I wouldn't mention `Functor` here, and just say "between type constructors" instead. `runCoyoneda` doesn't need this to be a `Functor`, for example.

src/Data/NaturalTransformation.purs
```diff
@@ -0,0 +1,14 @@
...
+-- | constraint is not enforced).
+-- |
+-- | A natural transformation allows a functor to be transformed into another
```
paf31: I'm not sure about the connection with `map`. I'd say something like "transforms a functor" and something about uniformity, and give an example like `listToMaybe`, which I think provides a decent intuition.

src/Data/NaturalTransformation.purs
paf31, line 19: 👍 Sounds good.

garyb:
Updated the description. I used `fromFoldable` as an example as we don't have `listToMaybe`.

I do think it's worth mentioning that our ~> does not require functor instances though, since technically it should if we're going to call it a natural transformation. I moved it to a note at the end rather than being part of the up front explanation instead though.

hdgarrood:
Does parametricity already ensure that, for a natural transformation `f`, we have `f >>> map g == map g >>> f`?

texastoland:
Is it not enforced because you might want a `NaturalTransformation` without that constraint or something else?

paf31:
👍 Looks good to me, thanks!

garyb:
@AppShipIt yeah, there are lots of things where the constraint would be unhelpful, Phil's coyoneda example being one. 

`Coyoneda f` gives you a `Functor` for any `f`, regardless of whether it has a `Functor` instance or not, so it would not be possible to write the type of `lowerCoyenda` as `Coyoneda f ~> f` without imposing `Functor f`, which kinda defeats the point.

I could be wrong here, so please someone jump in if my ignorance is showing, but I think any type of kind `* -> *` can be considered a lowercase-f-functor regardless of whether there is a `Functor` instance, because our `Functor` is more restrictive than that, as it's for endofunctors.

garyb:
Hmm, maybe not, as this is the natural transformation of endofunctors too, the category is fixed as `(->)`.

I don't know :wink:

texastoland:
Thanks @garyb 🙏

PR #83, opened 2016-06-04
Add generalized `flip`

### LINKED ISSUE #82: Functor function
I don't have a good name for this—nor do I care what the name is. Been calling it `flip'` in my head, since it generalizes `flip`. It's been lingering in `purescript-lens` for too long, and doesn't make sense there anymore. Should it be moved here?

``` purescript
infixl 1 flip' as ??

flip' :: forall f a b. (Functor f) => f (a -> b) -> a -> f b
flip' ff x = map (\f -> f x) ff
```

Some example uses:

``` purescript
longEnough :: String -> Bool
hasSymbol :: String -> Bool
hasDigit :: String -> Bool
password :: String

validate :: String -> List Bool
validate = flip' [longEnough, hasSymbol, hasDigit]
```

``` purescript
flip' (-) 3 4 -- => 1
```

``` purescript
foo <*> pure x == foo ?? x
```

joneshf:
Closes #82 

The name of this function along with its operator alias can change.
I don't know a better name for either and have no strong feelings.


texastoland:
Wow cool!

``` hs

instance ff ::             Functor (a -> _)
flip        :: ∀ a b c   .                     (a -> (b -> c)) -> b -> (a -> c)
flip'       :: ∀   b c f . Functor  f       =>  f    (b -> c)  -> b ->  f    c
flip'       :: ∀ a b c   . Functor (a -> _) =>  a -> (b -> c)  -> b -> (a -> c)
```

Do we need to keep the specialized version? Does it need an operator?

paf31:
Yes, we should keep the specialized (optimized) version, I think.

texastoland:
That's going down the Haskell `map` versus `fmap` road. In that case call it `fflip`. Rather than naming conventions for specialized/default variants of functions there should a way to provide implementations in type classes (e.g. `DefaultSignatures`) or without (as in this case). But adding TIMTOWTDI seems undesirable.

joneshf:
Don't let the name `flip'` sway your judgement. `const` and `pure` live in harmony. `compose` and `map` live in harmony.

Call it some other name if that makes it easier to deal with. `switch`? `surface`? `embed`? It takes a function in some computational context and returns a function that stuffs the result in the computational context.

Again, I don't care the name, so use what you like best.

> TIMTOWTDI

wat?

michaelficarra:
"There is more than one way to do it".

texastoland:
> `const` and `pure` live in harmony. `compose` and `map` live in harmony.

Excellent points. I still think it'd be better to address the user facing issue of specializing implementations than keep renaming things.

> <q>There is more than one way to do it.</q>

(Perl) cf. <q>There should be one—and preferably only one—obvious way to do it.</q> (Python)

paf31:
> "There is more than one way to do it".

This doesn't bother me so much.

joneshf:
What's the verdict here?

texastoland:
Sounds like it just needs a name. I'd like any function to be able to be specialized but it's mostly unrelated.

garyb:
`fapply` perhaps? Not a great name, but it's definitely closest in signature to `apply`, but only requires the `Functor` constraint. It seems more `apply`-y than `flip`-y to me anyway, even though when dealing with `(->) a` it is indeed flip.

garyb:
`flap`? :smile:

garyb:
^ in all seriousness I do actually like `flap`, it doesn't really mean anything (useful for an abstract function), and if you try really hard I guess it kind works:
1. it's part `flip`, part `apply`
2. it takes a function in a context, opens a flap and throws `a` in

texastoland:
It comes up in [Hayoo](http://hayoo.fh-wedel.de/?query=f+%28a+-%3E+b%29+-%3E+a+-%3E+f+b) but no solid convention.

texastoland:
`unapply`? It also looks vaguely (less than `apply`) like `unfoldr` and `sieve` in `Profunctor`.

paf31:
To me, one of the most useful places for this function is alongside `<$>` and `<*>`, since you can use it to apply a function when one of the arguments is not wrapped in an `f`, without using `pure`. 

So I suggest it should be called `applied`, with an infix alias of some kind.

epost:
`flap` = `a`bstract `flip`. I love it!

joneshf:
> an infix alias of some kind.

One that comes to mind is `(<@>)`. to pun off of @garyb, it's got an `a` inside a "flap":

> it takes a function in a context, opens a flap and throws `a` in

:wink:

joneshf:
So, we're at `flap` and `<@>`. Bueno, or different name?

paf31:
Sounds good, thanks! Do the precedence levels work out so that this can be used with `<*>`?

joneshf:
Do you have an example of how you'd like it to parse?

texastoland:
Perhaps it should be a test case?

paf31:
Something like `threeve <$> Just 1 <@> 'a' <*> Just true` I think.

paf31:
If the operator can be made to work like that, I think it would be nice to put something about it in the comment as well.

joneshf:
Updated. Didn't add a test since there's no precedent for what a ps test looks like here.

paf31:
Thanks, and sorry this took such a long time to merge!

joneshf:
No worries. Thanks for merging!

PR #84, opened 2016-06-07
Remove CommutativeRing constraint from Field

garyb:
This is already implied as `CommutativeRing` is a superclass of `EuclideanRing`. Correspondingly, this shouldn't be a breaking change, right?


paf31:
👍 Looks good, and yes, non-breaking, I think.

garyb:
Before releasing I compiled all the core libraries with the patch applied to double check, and there were no problems there at least. I think our any `Field` instances we have are already in the prelude though so it probably wouldn't have turned up much.

PR #85, opened 2016-07-09
Check javascript files with flow

ghost:
I don't know if this is really interesting or not but I decided on a whim to try adding [flow comments](https://flowtype.org/blog/2015/02/20/Flow-Comments.html) to the javascript files. Given that it works pretty smoothly and the changes are relatively non-invasive, maybe it would be worth using.

This PR just adds `flow-bin` to the development dependencies and modifies the `test` script to run flow along with the other tests. Since the flow annotations are only present in comments, no transformations with babel are needed. Other tools should just ignore them.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,0 +1,9 @@
+[ignore]
+.*/node_modules/.*
+
+[options]
+esproposal.class_instance_fields=enable
+esproposal.class_static_fields=enable
+esproposal.decorators=ignore
+module.use_strict=true
+suppress_comment=.*\\$FlowIgnore.*
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
     "test": "psc \"src/**/*.purs\" \"bower_components/purescript-*/**/*.purs\" \"test/**/*.purs\" && psc-bundle \"output/**/*.js\" --module Test.Main --main Test.Main | node"
   },
   "devDependencies": {
+    "flow-bin": "^0.28.0",
     "jscs": "^2.8.0",
     "jshint": "^2.9.1",
     "purescript-psa": "^0.3.8",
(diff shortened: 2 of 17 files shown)


paf31:
Interesting, thanks!

I think this makes the JS code more readable in some cases, like the `Ordering` example (even if the comments are a bit noisy in general).

👍 Looks good!

ghost:
Yeah, I find the comments generally make it easier to read and understand the javascript too, and of course having that part checked to some extent is also nice.

As far as readability goes, the Atom `language-babel` package supports highlighting in these comments now ([example](https://github.com/gandm/language-babel/pull/213)) so users of the Atom purescript could benefit from having that installed.

ghost:
@paf31 Come to think of it, how hard do you think it would be to actually generate annotations like this directly from the FFI declarations in the corresponding purescript code?

I realize there'd be some issues getting that to work perfectly due to the differences in the type systems but it seems like for a lot of simpler things that deal with basic javascript types it could be fully automated. You could just punt on the harder cases and fail, requiring that the user intervene and provide an explicit translation.

It would be interesting to be able to generate these annotations from the FFI declarations and run flow against the javascript code as part of the normal type-checking phase for purescript that interfaces with external code.

I guess maybe the easiest way to do this would be to implement the PS FFI -> flow annotations as some sort of babel transformation unless you already have a good way to do this sort of thing from within purescript.

Again, not sure if it's really worth it but it would be kind of neat if it worked.

paf31:
@freebroccolo This sounds quite similar to the idea from the `roadmap` repo, to produce and consume TypeScript `.d.ts` files. We have a similar problem there, in that the type systems don't quite match up, but we can do something like fall back to an abstract type like `Foreign` if an external type doesn't map nicely to the type system.

One thought I had was that it would be nice to generate template FFI files based on things like IDL files or `.d.ts` files, or even something like a YAML file containing a list of properties. I think adding Flow annotations would be a really natural fit for a tool like that.

It might be a good fit for the compiler too, but I think I'd rather leave it up to either the developer, or an external tool, to be honest.

texastoland:
Perhaps a marker interface pattern could be used for type classy things? But many common cases align 1:1.

paf31:
It's not just type classes though. Records, rank-n types are also problematic, and even things like functions, since they can have effects.

PR #93, opened 2016-09-26
Fix typo, document non-law-abiding numeric instances

### LINKED ISSUE #30: Number is a poorly behaved Num, Ord
By using the native IEEE-754 floats for `Number`, we inherit a bunch of unfortunate quirks. `+Infinity`, `-Infinity`, `+0`, `-0`, and `NaN` break a several of the `Ord` and `Num` laws. Loss of precision breaks a few more.

I don't think Purescript should give up on `Number` entirely--it's clearly useful in terms of performance, and FFI, and generating readable compiled code. Still, I think its an egregious enough issue to warrant some sort of treatment.

Some potential approaches:
1. Leave `Number` as is, but document it as a non-conforming typeclass instance.
2. Do (1) and also move `Number` out of purescript-prelude, and into purescript-number.
3. Do (2) and also introduce a well-behaved numeric type to the prelude (e.g. `Rational`).
4. Do (1) and also introduce a well-behaved numeric type to the prelude. Encourage use of new type over `Number` for serious mathematics.

### LINKED ISSUE #60: Arithmetic overflow
The documentation currently does not specify what happens on arithmetic overflow. If the behaviour is not meant to be undefined then it should be specified.

### LINKED ISSUE #89: Semiring docs typo
Annihiliation -> Annihilation

garyb:
Resolves #30, resolves #60, resolves #89.

### DIFF
--- a/src/Data/Eq.purs
+++ b/src/Data/Eq.purs
@@ -10,6 +10,11 @@ import Data.Void (Void)
 -- | - Reflexivity: `x == x = true`
 -- | - Symmetry: `x == y = y == x`
 -- | - Transitivity: if `x == y` and `y == z` then `x == z`
+-- |
+-- | **Note:** The `Number` type is not an entirely law abiding member of this
+-- | class due to the presence of `NaN`, since `NaN /= NaN`. Additionally,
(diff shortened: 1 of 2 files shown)


paf31:
👍

PR #94, opened 2016-09-26
Array semigroup optimisation

### LINKED ISSUE #63: Array Semigroup instance should check length
If either of the arrays are 0-length, then we should return the other as is instead of unnecessarily creating a copy.

garyb:
Resolves #63

Unless @hdgarrood still objects?

### DIFF
--- a/src/Data/Semigroup.js
+++ b/src/Data/Semigroup.js
@@ -10,6 +10,8 @@ exports.concatString = function (s1) {
 
 exports.concatArray = function (xs) {
   return function (ys) {
+    if (xs.length === 0) return ys;
+    if (ys.length === 0) return xs;
     return xs.concat(ys);
   };
 };


paf31:
👍 LGTM

hdgarrood:
My gut says that the previous code is probably going to be better on average but I'm not all that confident and I'm not going to find the time to test it properly, so sure, go ahead.

garyb:
Did you see my comment on the other issue? I'd be extremely surprised if checking array length is more expensive than shallow copying...

We can always revert if it turns out to be worse.

hdgarrood:
Oh yeah sure, I agree that this is clearly a win in the case where one of the argument is indeed empty. However, I think it's also quite clear that this will increase running time when neither argument is empty, so programs that do lots of concatenation with non-empty arrays will suffer.

My concern with `foldMap` was that given how comfortable and easy it is to use to find the 'sum' of some huge data structure of type `Foldable f => f (Array a)` using `foldMap`, assuming that it contains lots of nonempty arrays, this change could have a rather severe effect on code like that. Admittedly that sounds extremely specific after having actually written it down.

natefaubion:
I consider it the same as what you'd write for lists.
http://hackage.haskell.org/package/base-4.9.0.0/docs/src/GHC.Base.html#%2B%2B

garyb:
I just benchmarked something to check an assumption I have: I'm pretty sure arrays carry their `.length` as a static value, and do not compute it dynamically, so the overhead should be miniscule (the cost of a property read).

I constructed an empty array, and an array with `1e7` items, and then iterated reading the length `1e9` times for each. There was no perceptible difference between reading the length of either array.

hdgarrood:
Sure, it was also my belief that reading `length` is O(1). It wasn't that i was worried about complexity, i.e. I don't think this change makes any difference in the big O sense, it's just a small change in constant factors that I was guessing may become more pronounced in programs that do lots of concatenation.

PR #96, opened 2016-09-30
Add "whenM" and "unlessM"

garyb:
I find myself wanting to do something like this relatively often - having functions like these saves binding additional variables, which is particularly desirable considering how strict PS is about name shadowing.


paf31:
Sounds good. I find myself wanting these functions too.

PR #103, opened 2017-01-16
Add Field Boolean instance

hdgarrood:
This is known as F_2 or GF(2), see https://en.wikipedia.org/wiki/GF(2)

It also nicely fills a gap for logical xor, as there's no operator for
this as far as I know. This commit provides one in the form of `+`.

### DIFF
--- a/src/Data/CommutativeRing.purs
+++ b/src/Data/CommutativeRing.purs
@@ -20,3 +20,4 @@ class Ring a <= CommutativeRing a
 instance commutativeRingInt :: CommutativeRing Int
 instance commutativeRingNumber :: CommutativeRing Number
 instance commutativeRingUnit :: CommutativeRing Unit
+instance commutativeRingBoolean :: CommutativeRing Boolean
--- a/src/Data/EuclideanRing.purs
+++ b/src/Data/EuclideanRing.purs
@@ -40,6 +40,15 @@ instance euclideanRingUnit :: EuclideanRing Unit where
   div _ _ = unit
   mod _ _ = unit
 
+instance euclideanRingBoolean :: EuclideanRing Boolean where
+  degree false = 0
+  degree true = 1
+
+  div x true = x
+  div _ _ = false -- undefined?
+
+  mod _ _ = false
+
 foreign import intDegree :: Int -> Int
 foreign import intDiv :: Int -> Int -> Int
 foreign import intMod :: Int -> Int -> Int
--- a/src/Data/Field.purs
+++ b/src/Data/Field.purs
@@ -22,3 +22,4 @@ class EuclideanRing a <= Field a
 
 instance fieldNumber :: Field Number
 instance fieldUnit :: Field Unit
+instance fieldBoolean :: Field Boolean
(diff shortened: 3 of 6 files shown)


hdgarrood:
I think it might be better to use `unsafeCrash` for handling division by zero, actually. Is that going to be workable wrt dependencies?

garyb:
👍 from me. I think division by zero as being undefined (wrong :wink:) is probably better than raising an unadvertised runtime error, personally.

hdgarrood:
I'd much rather it raise an error. If any program ends up performing a division by zero at runtime, that almost always means that the programmer has made an error, and therefore if we produce any value that appears valid as a result, it's most likely going to be garbage which leads to things like bugs which manifest far from the original error, or you realise a few weeks later that some proportion of rows in your database are garbage and should be discarded (but you don't know which rows). It's for this reason that I try to avoid JS and PHP more than other dynamic languages, because JS and PHP like to keep going with nonsense values when the programmer has clearly made an error.

In an ideal world all uses of `div` would be accompanied by a proof that the divisor is nonzero, but in the absence of that, I think throwing an error is the best we can do.

For the same reason, it really bothers me that in PS, `1 / 0 == 0`. I think we should throw an error there too, actually.

garyb:
My problem with it is it runs counter to how the entire of the rest of the PS ecosystem works; not having runtime errors is one of the most important features of the language as far as I'm concerned. Unless they're advertised via the type system - `Eff` exceptions, `Partial`, etc.

I'd much rather add a `Partial` constraint to `div`, or require proof of nonzero-ness, or do just about anything other than letting it throw would be better. Unfortunately I think doing any of those things is not great either, as it would impose a performance penalty on native divide operations, and would be kinda irritating to use. It would undoubtedly infuriate some people who already (wrongly) think we're more interested in academic purity than real world usage also.

hdgarrood:
Actually, what about adding a `Partial` constraint and also adding an optimization pass that removes all nullary type class dictionaries?

hdgarrood:
I agree with everything you say, there's no obviously good solution, except that I think having programs continue running with known-garbage data is the most dangerous option.

garyb:
Disappearing nullary constraints would actually change the semantics of the language. @LiamGoodacre discovered this: http://ircbrowse.net/browse/purescript?id=436533&timestamp=1479836300#t1479836300 although I think people should come and take your keyboard away if you ever actually use that in real code 😜

hdgarrood:
Oh, haha, yes, I remember that. Yes, if we did this, I guess it would have to be a breaking change, although I think as long as we document it clearly in the release notes and preferably do it before 1.0 (if we're going to do it at all) I think that would be ok.

hdgarrood:
Actually sorry, I think I mean empty type class dictionaries rather than nullary type class dictionaries.

paf31:
`F_2` might be the only field with two elements, but it's not the only `Semiring`. There are at least two others which come from the boolean algebra structure. I think this should be a `newtype`.

hdgarrood:
Ok, good point, agreed. What are they, by the way? The only other one I can think of is (true, (&&), false, (||))

garyb:
The same but inverted I think.

paf31:
You can flip `&&` and `||` to get a second valid semiring since both distribute over each other.

PR #104, opened 2017-01-16
[BREAKING] Remove unit instances

### LINKED ISSUE #102: Remove EuclideanRing Unit instance
Apparently almost all texts regard the zero ring (i.e. the ring with a single element, which corresponds to our `Ring Unit` instance) as not being an integral domain: http://math.stackexchange.com/questions/1326505/is-the-zero-ring-a-domain, and for similar reasons it is generally agreed that there is no field with one element (see: https://en.wikipedia.org/wiki/Field_with_one_element).

The `EuclideanRing Unit` / `Field Unit` instances *do* satisfy all the laws we've specified for our type classes, though, AFAICT. So perhaps that means some laws need minor amendments to account for how `zero` behaves. I should be able to better say what that might be later in the semester once I know a bit more about rings and ideals and things. :)

hdgarrood:
Fixes #102. I also added a separate commit with a note that Euclidean rings and Euclidean domains are the same thing, as I wasn't sure of this the first time I came across this idea.

### DIFF
--- a/src/Data/EuclideanRing.purs
+++ b/src/Data/EuclideanRing.purs
@@ -8,14 +8,13 @@ module Data.EuclideanRing
 import Data.CommutativeRing (class CommutativeRing)
 import Data.Ring (class Ring, sub, (-))
 import Data.Semiring (class Semiring, add, mul, one, zero, (*), (+))
-import Data.Unit (Unit, unit)
 
 -- | The `EuclideanRing` class is for commutative rings that support division.
 -- |
 -- | Instances must satisfy the following law in addition to the `Ring`
 -- | laws:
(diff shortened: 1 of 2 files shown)


paf31:
Do you know why `one /= zero` might be required?

hdgarrood:
I don't think it's required in that the hierarchy ceases to be consistent or meaningful without it, but my impression is that not having this extra requirement means that what PureScript calls a Euclidean ring or a field differs from what mathematics calls a Euclidean ring or field, and that seems to me to be a problem worth fixing.

hdgarrood:
See #107 instead.

PR #105, opened 2017-01-17
Add short motivation of Void type

chexxor:
I came to Void to decide if I should use it for describing an FFI function: `enable() :: Boolean` - would the arg be `Unit` or `Void`? I had to ask on IRC how Void is intended to be used. Hopefully this short doc answers the question for others coming to it.

I'd like suggestions for better docs for this, also, btw.


hdgarrood (approved):
Looks good!

garyb:
Thanks!

Issue #99, opened 2016-11-04
More documentation for EuclideanRing


clayrat:
Is there any practical use to the `degree` function? One could probably use it to write a law test for `div` and `mod`, but beyond that?

garyb:
There probably are some uses that I'm unaware of, but it's primarily there so we have a way of talking about the laws for `div` and `mod`. Also it's part of the definition of what a Euclidean ring is, so dropping it from the class would be odd in that regard, if we name it as such.

We adopted `EuclideanRing` over the old `ModuloSemiring` to have a good way of expressing the fact that division by zero is undefined, since the `div` behaviour is only specified when `degree a > 0` and `degree a <= degree (a * b)`. We could have added some law(s) to the old `ModuloSemiring`, but that whole thing was pretty ad-hoc already, whereas at least now we're formulated in terms of a structure that has previously been studied!

Having it just because it's necessary for the laws is good enough reason to have it anyway, IMO!

clayrat:
I see. Might be worth adding this explanation in some form as a comment to the code!

hdgarrood:
Reading this, looking at the laws for `EuclideanRing` again, and comparing it to the definition of a euclidean domain, I'm a bit confused. My most important question is, do we agree with the definition given in https://en.wikipedia.org/wiki/Euclidean_domain, and is this what we're aiming to capture with the `EuclideanRing` class?

If the answer to the above is yes, then:
- How did we end up with this formulation of `EuclideanRing`? It looks a bit different from what I would have expected, given the Wikipedia definition. For example, I would have expected to see something like: "for any `a` and any `b` not equal to 0,  let `q = a / b`, and `r = a `mod` b`. Then `a = q*b + r`, and either `r == 0` or `degree r < degree b`". Or perhaps a slightly more sensible formulation would be to replace "either `r == 0` or `degree r < degree b`" with just "`degree r < degree b`", since the Wikipedia definition says that a Euclidean function does not actually have to be defined at 0, but in our case it's probably easier to have `degree` just take any `a` (including `zero :: a`), and probably we want `degree zero == 0`.
- If we are trying to deal with division by zero by saying that the `div`/`mod` behaviour is only specified when `degree a > 0` and `degree a <= degree (a*b)`, doesn't that admit a law-abiding (but wrong) instance for any `CommutativeRing` in which we have `degree = const 0`, and then we are free to do literally anything with `div` and `mod`?
[...]

PR #106, opened 2017-01-26
Add more instances of the form `C b => C (a -> b)`

garyb:
As pointed out by @rightfold.

### DIFF
--- a/src/Data/BooleanAlgebra.purs
+++ b/src/Data/BooleanAlgebra.purs
@@ -18,3 +18,4 @@ class HeytingAlgebra a <= BooleanAlgebra a
 
 instance booleanAlgebraBoolean :: BooleanAlgebra Boolean
 instance booleanAlgebraUnit :: BooleanAlgebra Unit
+instance booleanAlgebraFn :: BooleanAlgebra b => BooleanAlgebra (a -> b)
--- a/src/Data/CommutativeRing.purs
+++ b/src/Data/CommutativeRing.purs
@@ -20,3 +20,4 @@ class Ring a <= CommutativeRing a
 instance commutativeRingInt :: CommutativeRing Int
 instance commutativeRingNumber :: CommutativeRing Number
 instance commutativeRingUnit :: CommutativeRing Unit
+instance commutativeRingFn :: CommutativeRing b => CommutativeRing (a -> b)
--- a/src/Data/Ring.purs
+++ b/src/Data/Ring.purs
@@ -27,6 +27,9 @@ instance ringNumber :: Ring Number where
 instance ringUnit :: Ring Unit where
   sub _ _ = unit
 
+instance ringFn :: Ring b => Ring (a -> b) where
+  sub f g x = f x - g x
+
 -- | `negate x` can be used as a shorthand for `zero - x`.
 negate :: forall a. Ring a => a -> a
 negate a = zero - a
--- a/src/Data/Semiring.purs
+++ b/src/Data/Semiring.purs
@@ -44,6 +44,12 @@ instance semiringNumber :: Semiring Number where
   mul = numMul
   one = 1.0
 
+instance semiringFn :: Semiring b => Semiring (a -> b) where
+  add f g x = f x + g x
+  zero = \_ -> zero
+  mul f g x = f x * g x
+  one = \_ -> one
+
 instance semiringUnit :: Semiring Unit where
   add _ _ = unit
   zero = unit


garyb:
It's not possible up to `Field`, as `degree` can't be implemented for `EuclideanRing b => EuclideanRing (a -> b)`.

paf31:
👍

natefaubion:
> It's not possible up to Field, as degree can't be implemented for EuclideanRing b => EuclideanRing (a -> b).

To me this is one of the main problems with `degree`. No functors can implement this.

hdgarrood:
What are the other problems with `degree`? Also could you explain why no functors can implement it, I'm not sure I follow.

garyb:
Maybe there's some other formulation we can use that would avoid the need for `degree`. @hdgarrood has been looking at various aspects of it recently, like with #102, and had some points about it #99 too. I don't know enough about it without doing some research to say definitively right now.

garyb:
Oh, he's here :smile:

natefaubion:
The problem is the `a -> Int`. If I wanted a derived degree, based on the underling `a`, I would have to unwrap it first and then call `degree`. This only permits functors that can go from `f a -> a`. Since it appears that all the instances ignore their arguments, maybe it should be `Proxy a -> Int`?

hdgarrood:
Hehe, hi :) I think the laws need a bit of work but I'm generally in favour of the current hierarchy, although I could be persuaded otherwise.

Having `degree` allows us to write a generalised `gcd` function, which is quite nice. Arguably we should write that and export it from `Data.EuclideanRing`; I think that's basically the whole point of Euclidean rings, even.

hdgarrood:
If you wanted the EuclideanRing instance for polynomials, you would need to look at the value to determine its degree, so a proxy wouldn't be enough. I think Int is probably the simplest example of where you do need the value, actually.

garyb:
I'm not sure `Proxy (a -> b) -> Int` or `Proxy (F a) -> Int` would necessarily be any better anyway?

hdgarrood:
How would you go about deriving `degree`?

natefaubion:
I'm definitely out of my comfort zone when talking about details and laws surrounding the hierarchy. I just know that there is a limit to what you can implement in the numeric hierarchy for functors. I noticed this when I tried implementing numeric instances for `Future`. There's no reason one couldn't lift any numeric operation over it, but since I can't go from `Future a -> a` (by very definition of `Future`), I can't implement `degree`. So you can lift all the other numeric operations, but not`mod` and `div`.

paf31:
You can assume `degree` is `const 1` if you assume a `Field` for the codomain type.

natefaubion:
> You can assume degree is const 1 if you assume a Field for the codomain type.

I guess so, and that covers the usual operations, but I just think it's a somewhat arbitrary restriction by fact that you can't lift `EuclideanRing` over functors in general.

paf31:
Maybe that just means that functions are not a Euclidean domain in general.

natefaubion:
> Maybe that just means that functions are not a Euclidean domain in general.

At this point I have nothing to contribute 😆

hdgarrood:
I'm not sure a type like `Future` should have these instances; it probably won't satisfy eg distributivity if it's an effect-y type like `Aff`, `Future`, or `Eff`.

natefaubion:
I think it makes sense to lift division of some `f` _shrugs_.

natefaubion:
If you want another example, how about `Lazy`? While you can do `f a -> a`, it has operational implications that may not be so good.

hdgarrood:
Oh yeah, sure, no argument there. But I think that would be a different kind of division to what we currently have here. I think we basically have to choose between having mathematical operators conform to well-understood structures like rings and fields (well, well-understood by mathematicians at least) and being able to do the sort of thing you'd like to do with types like `Future`. I think both approaches are valid but I prefer the one we have now. Perhaps that's unsurprising given that I'm a maths student ;)

hdgarrood:
How would you feel about having a different set of lifted mathematical operators like:

```purescript
liftedAdd :: (Applicative f, Semiring a) => f a -> f a -> f a

infix 5 liftedAdd as ^+
```

Or something like that? They probably ought to go in a separate library rather than Prelude I guess.

With respect to `Lazy` it would of course be nice to be able to say "if we have a `Field`, don't force it, just return 1" (would instance chains let us say this?). In general, though, I think forcing it probably is the right thing if you do want the degree of something.

natefaubion:
> With respect to Lazy it would of course be nice to be able to say "if we have a Field, don't force it, just return 1" (would instance chains let us say this?). In general, though, I think forcing it probably is the right thing if you do want the degree of something.

The point was that you can't do `degree :: Lazy a -> Lazy Int`.

paf31:
> it probably won't satisfy eg distributivity if it's an effect-y type like

 Surely any `Applicative` is enough for the proof to go through?

hdgarrood:
If you were writing a `EuclideanRing a => EuclideanRing (Lazy a)` instance I think the `degree` implementation would have to be `degree <<< force`; that should be ok, right?

I think whether laws are satisfied does depend on the specific Applicative you have. I'm fairly sure these ones that have just been merged are fine, for instance.

paf31:
I'm not sure that is ok actually. Maybe in the special case of `Lazy` but not in general. Even with `Lazy` it seems a bit sketchy.

What I mean is that everything except `degree` can be implemented for `f a` when `f` is `Applicative` and `a` is a `EuclideanDomain`. But for `degree`, you need `a` to be a `Field`.

natefaubion:
You wouldn't be able to use `EucledianRing` over `Lazy` and guarantee that it preserves laziness.

hdgarrood:
> What I mean is that everything except degree can be implemented for f a when f is Applicative and a is a EuclideanDomain. But for degree, you need a to be a Field.

I think this is only true if you're willing to ignore laws, i.e. if you only care about whether these instances pass type checking?

> You wouldn't be able to use EucledianRing over Lazy and guarantee that it preserves laziness.

I don't see why this is a problem. In general you expect lazy values to perform computation as required to compute any portion of them which you ask for; asking for the `degree` of a `Lazy a` is one of the operations that does require forcing the entire thing. Nobody talks about how `length :: [a] -> Int` fails to preserve laziness in Haskell; I don't see why this is different.

paf31:
> I think this is only true if you're willing to ignore laws

How are the laws broken?

natefaubion:
Computing length is a thunk though. You can't write `someAlgorithim :: EuclideanRing a => ...` and guarantee that `someAlgorithm` won't force things to soon. The operator is always strict, and it's not possible to move it under a thunk unless you code specifically to `Lazy`.

hdgarrood:
For instance:

*edit: pun not intended*

```purescript
instance semiringEff :: Semiring a => Semiring (Eff eff a) where
  add = liftA2 add
  zero = pure zero
  mul = liftA2 mul
  one = pure one

foo :: Eff _ Int
foo = log "hi" *> pure 5

main = do
  -- By distributivity, these two things should be equal
  foo * (pure 1 + pure 2)
  (foo * pure 1) + (foo * pure 2)
```

hdgarrood:
I see what you mean about `Lazy` now, thanks. Is there a problem with using `map degree` if you want that behaviour?

paf31:
I see your point about `Eff`, thanks for elaborating 😄

natefaubion:
Wouldn't `Future` preserve distributivity due to its sharing?

natefaubion:
Or I should say, `purescript-aff-future` has effect sharing. Pulling from a Future at multiple points does not yield multiple effects.

hdgarrood:
Oops, yes, I think you're right. Although I suspect there are other laws that might cause trouble when we're dealing with effects - commutativity of `add` might be another suspect, for instance.

hdgarrood:
Or better, `a - a = zero` - presumably this won't be true if `a` causes an effect.

natefaubion:
> I see what you mean about Lazy now, thanks. Is there a problem with using map degree if you want that behaviour?

That would a different interface `someAlgo :: (EuclideanRing a, Functor f) =>`

> Or better, a - a = zero - presumably this won't be true if a causes an effect.

`a` doesn't cause an effect though. There is no function `Aff eff a -> Future a`, only `Aff eff a -> Aff eff (Future a)`, so presumably the effect has already taken place. `Future` should therefor be a "pure" interface.

hdgarrood:
The `Future` docs seem to indicate that the effects are only performed when the value is requested (e.g. via `wait`), and that does seem to be how it behaves, too:

```purescript
module Main where

import Control.Apply (lift2)
import Prelude
import Control.Monad.Aff
import Control.Monad.Aff.Console
import Control.Monad.Aff.Future

main = launchAff do
  a <- defer (log "hi" *> pure 2)

  -- By the Ring laws, assuming the proposed Ring Future instance, these should
  -- be equivalent
  let liftedSub = lift2 sub a a
  let liftedZero = pure (zero :: Int)

  log "sub"
  wait liftedSub >>= logShow
  log "zero"
  wait liftedZero >>= logShow
```

prints:

```
sub
hi
0
zero
0
```

So with the proposed `Ring Future` instance, `a - a` could trigger an effect but `zero` won't, i.e. in general, `a - a /= zero`.

hdgarrood:
A `Semiring Future` instance would violate the annihilation law, `a * zero = zero`, in a similar way: the LHS might perform effects but the RHS won't.

natefaubion:
That's fair. According to the type system, the effect has already occured (just suspended indefinitely), so I think it fulfills the letter of the law, but perhaps not the spirit 😄. In that case an eagerly forked `Future` _would_ fulfill the laws (that's how it originally operated anyway).

hdgarrood:
I think all we can say for certain is that an eagerly forked `Future` won't fail to satisfy the laws for this specific reason! But yes, I think an eagerly forked `Future` probably would be ok. It intuitively feels that way to me, at least.

I do think `degree` has the right signature, though. The point of it is to ensure that you can keep splitting something up into quotients and remainders, in a process a little bit like long division, and the remainders keep getting smaller. This essentially gives you a way of implementing Euclid's `gcd` algorithm and knowing that it won't diverge.

On the other hand, the class would still work without `degree` as a real member: you only need to know that such a function exists to implement `gcd`, you don't actually need to use it. But having to write it out when you define an instance is arguably a good thing from a documentation perspective, as e.g. it tells you about which instance you have for types that admit more than one instance, like `Number`.

My general view is that `lift2 add` is in general a rather different beast from `add` so I think it's a good thing to keep them separate. `lift2 add` is not that much effort to write out, and it's easy to lift any function which uses a general `Ring a` or `Field a` or something to work on `f a` by using what `Applicative f` gives you. So I'm not too bothered by a hierarchy that doesn't really encourage functor-y instances. But I accept that this is not a universal view.

hdgarrood:
So I think what I previously said about `degree` was actually incorrect: I'm not aware of any examples where you need to ask for the `degree` of something.

hdgarrood:
One more thing I've just realised: perhaps it's a good thing that the current hierarchy doesn't permit a `EuclideanRing a => EuclideanRing (r -> a)` instance. In general, you expect that division is fine as long as the denominator is nonzero. But that isn't enough for functions!

Note that "nonzero" for functions means "not equal to the function `const zero`". However when dividing functions using `lift2 div`, you need the image of the divisor to not contain zero in order to be safe; this is a much stronger condition than it just being nonzero.

PR #107, opened 2017-01-27
Rework EuclideanRing laws, fixes #99, #102

### LINKED ISSUE #99: More documentation for EuclideanRing

hdgarrood:
This change that fixes the EuclideanRing laws so that they do actually
describe a euclidean ring (I have discussed some problems with the
current set of laws in #99).

This is technically a breaking change, since a) the laws have changed,
and b) fixing the laws requires tightening them so that the
`EuclideanRing Unit` instance is no longer valid, and likewise also the
`Field Unit` instance. It's unlikely to break very much code, though.

---

This also subsumes my other PR, #104 - since the law changes are 'technically breaking but rather unlikely to actually break anything', just like the removal of the `Unit` instances, I thought it would make more sense to put these things together.

### DIFF
--- a/src/Data/EuclideanRing.purs
+++ b/src/Data/EuclideanRing.purs
@@ -11,13 +11,37 @@ import Data.Semiring (class Semiring, add, mul, one, zero, (*), (+))
 import Data.Unit (Unit, unit)
 
 -- | The `EuclideanRing` class is for commutative rings that support division.
+-- | The mathematical structure this class is based on is sometimes also called
+-- | a *Euclidean domain*.
 -- |
--- | Instances must satisfy the following law in addition to the `Ring`
+-- | Instances must satisfy the following laws in addition to the `Ring`
 -- | laws:
 -- |
--- | - Integral domain: `a /= 0` and `b /= 0` implies `a * b /= 0`
--- | - Multiplicative Euclidean function: ``a = (a / b) * b + (a `mod` b)``
--- |   where `degree a > 0` and `degree a <= degree (a * b)`
+-- | - Integral domain: `one /= zero`, and if `a` and `b` are both nonzero then
+-- |   so is their product `a * b`
+-- | - Euclidean function `degree`:
+-- |   - Nonnegativity: For all nonzero `a`, `degree a >= 0`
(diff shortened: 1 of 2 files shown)


src/Data/EuclideanRing.purs
```diff
@@ -8,16 +8,27 @@ module Data.EuclideanRing
...
+-- |   so is their product `a * b`.
+-- | - Euclidean function `degree`:
+-- |   - For all `a` and `b`, where `b /= zero`, let `q = a / b` and
+-- |     ``r = a `mod` b``. Then `a = q*b + r`, and also either `r = zero` or
+-- |     `degree r < degree b`.
+-- | - Submultiplicative euclidean function:
+-- |   - For all nonzero `a` and `b`, `degree a <= degree (a * b)`.
+-- |
+-- | For any `EuclideanRing` which is also a `Field`, one valid choice
+-- | for `degree` is simply `const one`. In fact, unless there's a specific
```
garyb: Minorest of nitpicks, could say `const 1` here since `degree` returns an `Int` rather than an `a`.

hdgarrood: Ah, yes, thanks! I thought I'd caught all of these already.

paf31:
👍

hdgarrood:
I think this might be one example of an extremely rare situation in which we technically have a breaking change but it would be better to release as a patch level bump. What do we think?

paf31:
What about using the new `Warn` class?

hdgarrood:
I think that would count as a breaking change since it would cause a compile error with older compilers that don't have Warn?

Although that's a good idea, we could change the laws and leave the Unit instances in, perhaps with a note that they should not be used, then remove them at the next major bump?

paf31:
Yes, I think that's probably best. After the next round of breaking changes, we can start to use `Warn`.

hdgarrood:
Ok cool, I've left the `Unit` instances in with a note that they will be removed. If you think this looks good I'll rebase (so that there's just one commit) and merge.

paf31:
👍 LGTM, thanks!

PR #108, opened 2017-01-27
Fix markup in Field law

hdgarrood:
I've checked that both Cheapskate (which Pursuit uses) and GitHub agree on how to render this, and that it agrees with what we want :)


hdgarrood:
Oh, and I also changed the law to use the polymorphic `zero` rather than the Int literal `0`.

paf31:
👍

PR #109, opened 2017-01-29
Add gcd and lcm

hdgarrood:
The EuclideanRing laws ensure that this terminates.


paf31:
👍 LGTM, thanks!

hdgarrood:
Oh actually, do you think I should add `lcm` while I'm at it?

```purescript
lcm a b =
  if a == zero && b == zero
    then zero
    else a * b / gcd a b
```

paf31:
Sure, makes sense to have both.

src/Data/EuclideanRing.purs
```diff
@@ -69,3 +73,17 @@ foreign import intDiv :: Int -> Int -> Int
...
+gcd :: forall a. (Eq a, EuclideanRing a) => a -> a -> a
+gcd a b =
+  if b == zero
+    then a
+    else gcd b (a `mod` b)
+
+-- | The *least common multiple* of two values.
+lcm :: forall a. (Eq a, EuclideanRing a) => a -> a -> a
+lcm a b =
+  if a == zero && b == zero
```
paf31: Should/can this be _or_ instead of _and_?

hdgarrood: I think it can, and actually it sounds like that would be better, according to the first paragraph of https://en.wikipedia.org/wiki/Least_common_multiple:

> In arithmetic and number theory, the least common multiple, lowest common multiple, or smallest common multiple of two integers a and b, usually denoted by LCM(a, b), is the smallest positive integer that is divisible by both a and b.[1] Since division of integers by zero is undefined, this definition has meaning only if a and b are both different from zero.[2] However, some authors define lcm(a,0) as 0 for all a, which is the result of taking the lcm to be the least upper bound in the lattice of divisibility.

paf31:
👍 🚢

PR #111, opened 2017-01-29
Add law functions

### LINKED ISSUE #110: Export law functions?
`quickcheck-laws` is really handy, but it's not usable in every situation. For example, I have a Polynomial type in [`polynomials`](https://github.com/hdgarrood/purescript-polynomials) which can only be given an `Arbitrary` instance if the type argument has `Eq`, `Semiring`, and `Arbitrary` instances. This means that I can't e.g. use `quickcheck-laws` to check the `Functor` laws with it, because that would require an `Arbitrary (Polynomial A)` instance, which in turn requires a `Semiring A` instance, which does not exist (`A` is defined in `quickcheck-laws`).

Also, I might want to use testing libraries other than quickcheck, e.g. strongcheck or jack.

I think we can make it easier to test that instances abide by the relevant laws with property tests in more scenarios by creating a few more modules in `Prelude`, perhaps one `.Laws` module accompanying each module which defines a type class. So for example `Data.Eq.Laws` could export `symmetric :: forall a. (Eq a) => a -> a -> Boolean` which would be nothing more than `\x y -> (x == y) == (y == x)`. We might then re-export all of these law functions from another module `Prelude.Laws` for convenience. All existing modules would be unaffected.

hdgarrood:
WIP. This will fix #110 when it's done.

I'd appreciate a quick review now just to check that we agree on this style before I go ahead and do all the others in the same way.

### DIFF
--- a/src/Control/Applicative/Laws.purs
+++ b/src/Control/Applicative/Laws.purs
@@ -1,0 +1,34 @@
+module Control.Applicative.Laws where
+
+import Control.Applicative
+import Control.Category (id, (<<<))
+import Data.Function (($))
+import Data.Eq (class Eq, (==))
+
+-- | Checks the identity law for a given value:
+-- |
+-- |     (pure id) <*> v = v
+identity :: forall f a. (Applicative f, Eq (f a)) => f a -> Boolean
+identity v =
+  (pure id) <*> v == v
+
+-- | Checks the composition law for a given triple of values:
+-- |
+-- |     pure (<<<) <*> f <*> g <*> h = f <*> (g <*> h)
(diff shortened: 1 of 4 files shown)


garyb:
Nice 👍

hdgarrood:
Come to think of it, I wonder if a better formulation would be returning a `{ lhs :: a, rhs :: a }` for the appropriate type `a`? That way, you're free to use some other form of equality checking. E.g. if both sides evaluate to functions of some kind, you might also generate a number of inputs and check those functions agree for all the inputs. Or if `a` was `Number`, you could use an approximate equality function.

hdgarrood:
I'm going to close this for now since it's not had any progress in almost two years now (!)

PR #114, opened 2017-02-10
Fix intDegree on edge cases

### LINKED ISSUE #113: degree bottom > top
`degree` on integers returns an out-of-bounds integer.

no-longer-on-githu-b:
Fixes #113.

### DIFF
--- a/src/Data/EuclideanRing.js
+++ b/src/Data/EuclideanRing.js
@@ -1,7 +1,7 @@
 "use strict";
 
 exports.intDegree = function (x) {
-  return Math.abs(x);
+  return Math.abs(x) | 0;
 };
 
 exports.intDiv = function (x) {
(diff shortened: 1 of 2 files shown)


hdgarrood (approved):

garyb:
I wonder if this should be clamped rather than modulo, due to the intended behaviour of `degree`. @hdgarrood?

hdgarrood:
Ah good point. Do we have `degree (bottom :: Int) == bottom` with this PR?

no-longer-on-githu-b:
I have no clue, but with this PR yes, `degree bottom == bottom`.

Sent from my iPhone.

hdgarrood:
I think @garyb is right, `degree bottom == top` would be better I think.

no-longer-on-githu-b:
Alright, changed that.

hdgarrood (approved):

PR #115, opened 2017-02-11
Add Discard class

paf31:
To support https://github.com/purescript/purescript/issues/1803

I tried a few different approaches, and this one seems nicest. It has the additional benefit that you can now use `do` notation with an arbitrary `Apply`, as long as you discard the results.

### DIFF
--- a/src/Control/Bind.purs
+++ b/src/Control/Bind.purs
@@ -74,10 +74,10 @@ foreign import arrayBind :: forall a b. Array a -> (a -> Array b) -> Array b
 -- | An example is the `Unit` type, since there is only one
 -- | possible value which can be returned.
 class Discard a where
-  discard :: forall f b. Apply f => f a -> (a -> f b) -> f b
+  discard :: forall f b. Bind f => f a -> (a -> f b) -> f b
 
 instance discardUnit :: Discard Unit where
-  discard u f = u *> f unit
+  discard = bind
 
 -- | Collapse two applications of a monadic type constructor into one.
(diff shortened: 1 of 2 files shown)


src/Control/Bind.purs
```diff
@@ -66,6 +68,17 @@ instance bindArray :: Bind Array where
...
+-- | A class for types whose values can safely be discarded
+-- | in a `do` notation block.
+-- |
+-- | An example is the `Unit` type, since there is only one
+-- | possible value which can be returned.
+class Discard a where
+  discard :: forall f b. Apply f => f a -> (a -> f b) -> f b
+
+instance discardUnit :: Discard Unit where
+  discard u f = u *> f unit
```
paf31: Maybe this needs to use `>>=` actually, if we don't want it to always strictly evaluate `f`.

garyb: Good point!

no-longer-on-githu-b:
Void instance?

paf31:
Hmm, I'm not sure about `Void`.

garyb (approved):

garyb (changes requested):
Actually, looks like there's an unused `unit` import :smile:

paf31:
Fixed, thanks.

garyb (approved):

paf31:
Do we want to merge some other PRs before making a release?

hdgarrood:
I think #112 and #114 would be good to merge before making a release. #112 is just a docs update, and #114 is a bug fix. #111 is still a way from being mergeable though.

PR #116, opened 2017-02-16
Bounded instance for Number

matthewleon:
I wasn't able to find any discussion of this omission, but I can imagine that there was some reason for it. In that case, at least it can be documented with this PR.

### DIFF
--- a/src/Data/Bounded.js
+++ b/src/Data/Bounded.js
@@ -3,5 +3,8 @@
 exports.topInt = 2147483647;
 exports.bottomInt = -2147483648;
 
+exports.topNumber = Number.MAX_VALUE;
+exports.bottomNumber = Number.MIN_VALUE;
+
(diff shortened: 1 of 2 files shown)


garyb:
Unfortunately this doesn't satisfy the law for `Bounded`, as `Infinity > top`... I guess `Number` is not good when it comes to obeying the laws generally anyway, but this still doesn't seem quite right to me.

hdgarrood:
Yeah, I'm not really sure what the best approach is here, but since `Number` is meant to represent the real numbers (albeit poorly) I am not sure that this instance is desirable. Do you have a use case in mind for it?

matthewleon:
Good points above. I had in mind some functions that accepted `Bounded` parameters but simply adding the desired bounds as parameters to a more general version works fine. I'm totally happy to see this issue closed off.

PR #118, opened 2017-03-09
Update Functor docs to use `map`

hdgarrood:
Now that the type class defines `map` and not `<$>`, this makes more
sense. It also fixes a problem where the previous docs used the old
operator section syntax, which was removed a while ago.


paf31:
👍 Looks good

PR #119, opened 2017-03-12
Rename discard

garyb:
@paf31 I renamed the class too, but I'm not sure it's any better :smile:... thoughts?


src/Control/Bind.purs
```diff
@@ -73,11 +73,11 @@ foreign import arrayBind :: forall a b. Array a -> (a -> Array b) -> Array b
 -- |
 -- | An example is the `Unit` type, since there is only one
 -- | possible value which can be returned.
-class Discard a where
-  discard :: forall f b. Bind f => f a -> (a -> f b) -> f b
+class BindDiscard a where
+  bind_ :: forall f b. Bind f => f a -> (a -> f b) -> f b
```
paf31, line 77: One thing - I might expect `bind_` to have type `Bind f => f a -> (a -> f b) -> f Unit` since the `_` usually refers to the return type.

garyb: Yeah, I wondered about that too. I'll make the change!

paf31: Which change?

paf31: No, it can't have that type, the `b` is the final result of the `do` block. I was just saying that `bind_` suggests that to me, so it might be surprising.

garyb: Ah right, oops.

garyb: `_bind`? 😆 

`voidBind`, similar to `voidLeft` / `voidRight` for `<$` and `$>`?

paf31: Honestly, I prefer `discard` the more I think about it :)

`ignore` or `ignoring` is another option.

garyb: I'm thinking about it in context of rebindable syntax, without the word `bind` in there it probably won't be too obvious what it relates to.

paf31: Yeah, `bind_` is fine, I'll not bikeshed any more :)

garyb: Well, I started it! :wink: I agree that `bind_` doesn't quite seem right after all, since it's not giving back a `Unit` result like the other cases; that's down to a misunderstanding I had of how it was working.

Re: rebindable syntax, would one even need to redefine `bind_`/`discard`/whatever if overriding `bind`? Or only if they were doing something "fancy"?

paf31: I think users would override `bind` and `discard` together, and they'd only need to override `discard` if they were discarding results, obviously.

PR #130, opened 2017-06-27
Re-add DivisionRing, resolves #128

### LINKED ISSUE #128: Re-add DivisionRing?
I am working on a quaternion library and I was wondering if it might make sense to add a `DivisionRing` type class again, as follows:

```purescript
class Ring a <= DivisionRing a where
  recip :: a -> a
```

with the law that `x * recip x = recip x * x = one` for all nonzero `x`. There's no requirement for the ring to be commutative, and so this would allow us to provide a `DivisionRing Quaternion` instance.

If we were going to do this, we should change `Field` so that `DivisionRing` is a subclass as well.

Now of course with any `EuclideanRing` you can simply write `one / x` for `recip x`. Also, admittedly, the only 'useful' instance I can think of is `Quaternion`; I did some googling and wasn't really able to find any other useful examples of non-commutative division rings. But I think it would be nice to have a way of asking for the multiplicative inverse of an element of an arbitrary division ring without requiring it to be commutative.

hdgarrood:
I also added brackets around the word "commutative" in the description
of the field class so that we can be consistent in terminology (i.e.
using the scheme where "field" = commutative division ring, "skew field"
= noncommutative division ring).

I thought it might be confusing to re-export `leftDiv` and `rightDiv`
from `Prelude`; since they're probably only going to be used very rarely
I thought it would be better to have people import them from
`Data.DivisionRing` if they want to use them.

Also I realised that we have the word "nonzero" with and without a hyphen in various places in this repo. Perhaps we should standardise on one?

### DIFF
--- a/src/Data/DivisionRing.purs
+++ b/src/Data/DivisionRing.purs
@@ -1,0 +1,55 @@
+module Data.DivisionRing
+  ( class DivisionRing
+  , recip
+  , leftDiv
+  , rightDiv
+  , module Data.Ring
+  , module Data.Semiring
+  ) where
+
+import Data.Ring (class Ring, negate, sub)
+import Data.Semiring (class Semiring, add, mul, one, zero, (*), (+))
+import Data.EuclideanRing ((/))
+
+-- | The `DivisionRing` class is for non-zero rings in which every non-zero
+-- | element has a multiplicative inverse. Division rings are sometimes also
+-- | called *skew fields*.
+-- |
+-- | Instances must satisfy the following laws in addition to the `Ring` laws:
+-- |
+-- | - Non-zero ring: `one /= zero`
+-- | - Non-zero multiplicative inverse: `recip a * a = a * recip a = one` for
+-- |   all non-zero `a`
+-- |
+-- | The result of `recip zero` is left undefined; individual instances may
+-- | choose how to handle this case.
+-- |
+-- | If a type has both `DivisionRing` and `CommutativeRing` instances, then
+-- | it is a field and should have a `Field` instance.
+class Ring a <= DivisionRing a where
+  recip :: a -> a
+
+-- | Left division, defined as `leftDiv a b = recip b * a`. Left and right
+-- | division are distinct in this module because a `DivisionRing` is not
+-- | necessarily commutative.
+-- |
+-- | If the type `a` is also a `EuclideanRing`, then this function is
+-- | equivalent to `div` from the `EuclideanRing` class. When working
+-- | abstractly, `div` should generally be preferred, unless you know that you
+-- | need your code to work with noncommutative rings.
+leftDiv :: forall a. DivisionRing a => a -> a -> a
+leftDiv a b = recip b * a
+
+-- | Right division, defined as `rightDiv a b = a * recip b`. Left and right
+-- | division are distinct in this module because a `DivisionRing` is not
+-- | necessarily commutative.
+-- |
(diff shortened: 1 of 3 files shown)


src/Data/DivisionRing.purs
```diff
@@ -0,0 +1,55 @@
...
+-- | - Non-zero multiplicative inverse: `recip a * a = a * recip a = one` for
+-- |   all non-zero `a`
+-- |
+-- | The result of `recip zero` is left undefined; individual instances may
+-- | choose how to handle this case.
+-- |
+-- | If a type has both `DivisionRing` and `CommutativeRing` instances, then
+-- | it is a field and should have a `Field` instance.
+class Ring a <= DivisionRing a where
+  recip :: a -> a
```
paf31, line 30: Just wondering - have you considered an API in terms of `leftDiv` and `rightDiv` instead?

paf31:
👍 Looks good to me! @garyb, any thoughts?

src/Data/DivisionRing.purs
hdgarrood: I haven't actually, but now that you mention it: I think if we had  `leftDiv` and `rightDiv` as class members, then `recip` would have to be defined as either `recip x = leftDiv one x` or `recip x = rightDiv one x`, and we'd have a risk of having the two not match up properly. Also I imagine it would be easy to mix them up if people had to declare their own instances! I think `recip` is a bit easier to explain too.

paf31: Good point, thanks.

hdgarrood:
Also, I've avoided making any breaking changes here, but for the future: this makes the `Field` class kind of redundant as a `Field` is precisely a commutative division ring. We could simply remove it, but it's nice to be able to write `Field a => [...]` rather than `DivisionRing a => CommutativeRing a => [...]` and I imagine a fair bit of existing code uses `Field` constraints too.

To address this I am aware of two options: we could potentially use that trick I suggested in #128:

```purescript
```

and nobody then needs to define a `Field` instance but `Field` constraints still work. Alternatively, once we have constraint kinds, would it be possible to use a type synonym so that `Field a => [...]` is the same as `DivisionRing a => CommutativeRing a => [...]`?

paf31:
> once we have constraint kinds, would it be possible to use a type synonym so that Field a => [...] is the same as DivisionRing a => CommutativeRing a => [...]?

No, unfortunately not, at least not with the way I'm implementing them at the moment. GHC uses a bit of a trick to do this, overloading tuple notation, but I don't particularly want to go that route.

hdgarrood:
Right, makes sense. Another approach I've seen, used by [Data.Group.Abelian](https://pursuit.purescript.org/packages/purescript-group/3.1.0/docs/Data.Group#t:Abelian), would be to define `type Field a b = CommutativeRing a => DivisionRing a => b`, so people could write `Field a b` instead of `Field a => b`. But I don't like that quite as much, personally.

paf31:
Once we have constraint kinds, we'll at least be able to write `class Both (c1 :: Constraint) (c2 :: Constraint)` once and for all :).

garyb:
This all looks good to me 👍 🎓

hdgarrood:
Fab, thank you both! Would you like me to make a release with a minor bump then?

paf31:
Yes please

hdgarrood:
Okay, v3.1.0 is tagged and on Pursuit :)

PR #135, opened 2017-08-18
Add `divisionringFunction` instance

joneshf:
From a discussion on slack.

### DIFF
--- a/src/Data/DivisionRing.purs
+++ b/src/Data/DivisionRing.purs
@@ -53,3 +53,6 @@ rightDiv a b = a * recip b
 
 instance divisionringNumber :: DivisionRing Number where
   recip x = 1.0 / x
+
+instance divisionringFunction :: DivisionRing b => DivisionRing (a -> b) where
+  recip f x = recip (f x)


garyb (approved):
LGTM 👍

hdgarrood:
This seems plausible but I'd like to try to prove it first.

hdgarrood:
In fact this seems non-obvious enough that we ought to have a proof in the comments, IMO. Give me a half hour or so, I think I can either come up with a proof or a counterexample.

hdgarrood:
Ok yeah I'm happy with this:

## `one /= zero`

since `b` is a `DivisionRing`, we must have `one :: b /= zero :: b`. Now `one :: a -> b` is defined as `one = \_ -> one`, and likewise `zero :: a -> b` is `zero = \_ -> zero`, so clearly `one :: a -> b /= zero :: a -> b`.

## `recip x * x = x * recip x = one`

Let `x :: a -> b`. Then `recip x * x = \y -> recip (x y) * (x y)`, which is the same as `\y -> one` by the `DivisionRing` laws since `b` is a `DivisionRing`. This is then the same as `\_ -> one` i.e. `one :: a -> b`. The same argument can be used to establish that `x * recip x = one`.

hdgarrood:
Oh wait no, sorry, this is actually incorrect, for the same reason that you can't have a `EuclideanRing b => EuclideanRing (a -> b)` instance! Suppose we do have this instance and consider:

```purescript
f :: Int -> Int
f 32 = 1
f _ = 0
```

Now this `f` is not the same as `zero :: Int -> Int`, since `zero :: Int -> Int` is just `const zero`. Since `f` is nonzero we should be able to ask for `recip f`. But if we apply `recip f` to anything other than `32`, the result is undefined. Therefore, we can't guarantee that `recip f * f = one`.

See also: [Why there can be no `EuclideanRing b => EuclideanRing (a -> b)` instance](https://gist.github.com/hdgarrood/2aa65073acf7fbdd9736780d552a8ad4)

joneshf:
I think I get it. Thanks for the review!

PR #136, opened 2017-08-26
Add Module

no-longer-on-githu-b:


src/Data/Module.purs
```diff
@@ -0,0 +1,50 @@
...
+-- |   - Associativity: `(a $+ b) $+ c = a $+ (b $+ c)`
+-- |   - Identity: `mlzero $+ a = a $+ mlzero = a`
+-- |   - Commutative: `a $+ b = b $+ a`
+-- | - Multiplicative identity: `one $* a = a`
+-- | - Multiplication distributes over addition:
+-- |   -  distributivity: `a $* (b $+ c) = (a $* b) $+ (a $* c)`
+-- |   - Right distributivity: `(a + b) $* c = (a $* c) $+ (b $* c)`
+-- | - Annihilation: `zero $* a = mlzero`
+-- | - Compatibility: `(a * b) $* c = a $* (b $* c)`
+class (Ring s) <= Module m s | m -> s where
```
paf31: `CommutativeRing`?

no-longer-on-githu-b: Yes, good catch.

hdgarrood:
The definitions I've seen only require a ring, not necessarily a commutative one.

As was pointed out in Slack, we ought really to have `AbelianGroup => Group => Monoid` and `class (AbelianGroup v, Ring s) <= Module v s`. So I think we need to either move parts of `monoid` in here, or decide that `Module` should be in a different core library instead. I'm tempted to say the latter, since I don't expect modules to pop up all that often.

Also there is https://github.com/morganthomas/purescript-vector-space - maybe we should ask if we could move that to core, split off a Module class, etc?

paf31:
The problem is that modules over non-commutative rings need both left and right modules. That seems to me like something rare enough that the commutative case (where the two coincide) is good enough for now.

We discussed on Slack that it might be inconvenient to have `Group` as a superclass, since then you can't have `Number` be a module over itself, for example, you need `Additive` or whatever. I'd be fine with this formulation, with newtype wrappers providing the monoid and group instances.

hdgarrood:
Would we need both left and right modules? As far as I can tell the only reason mathematicians don't just call left modules "modules" and not bother having right modules at all is syntactic convenience.

I think using that formulation and putting the instances on `Additive` is a bit nicer actually: the class definition and laws are then much simpler. I also think it's nice to only need to define the "any ring is a module over itself" instance once rather than n times.

hdgarrood:
Oh wait I see now: left and right modules over a non-commutative ring are actually slightly different because of the compatibility law.

Thanks for having the patience to rehash all of this despite me having been in Slack just a couple of hours ago and apparently having forgotten it all!

paf31:
Well, left and right modules do differ. For example, a ring is a module over itself, but the left and right modules would be different if the ring is not commutative.

paf31:
Haha, I saw your comment appear as I submitted mine :)

no-longer-on-githu-b:
I have added a `VectorSpace` subclass with a division method because otherwise you need to take the reciprocal of the scalar if you want to divide, which may be inefficient or imprecise.

hdgarrood:
Have we considered having just one `Module` class which is a left module over a not-necessarily-commutative ring? I think this would be nice as it adds a bit more generality without being more complex, and I think the terminology then matches up a bit better with mathematics too. I don't think it's ideal that people might see that we have a `Module` class, come across the [Wikipedia page titled "Module (mathematics)"](https://en.wikipedia.org/wiki/Module_(mathematics)), and then see that these things are subtly different.

Also, @rightfold, could you please remind me how you feel about the formulation with the `Group` superclass? The only thing that slightly bothers me about that is that `v <> w` doesn't look like addition, but we could perhaps provide an additional function `madd :: forall m r. Module m r => m -> m -> m`, which is just `append`, but with a stronger constraint and with a new operator alias that does look more like addition.

no-longer-on-githu-b:
@hdgarrood For the same reason `Semiring` doesn't have `Monoid` as a superclass. Types that are instances could have multiple possible `Group` instances, with no best one, so we choose to have no `Group` instance, and rely on newtypes to disambiguate.

hdgarrood:
Can you give an example of a module whose elements can be made into a group in more than one way, with no obvious "default" group structure?

no-longer-on-githu-b:
For example, `Number`:
R groups

I also feel rather strange about having a `Group Milliseconds` instance, where `append` does addition. But this may not be justified.

hdgarrood:
I'm not totally sold on that example, personally, because for the second one, the underlying set is really the reals without 0.

no-longer-on-githu-b:
Hmm, right. :) But, before you can have a `Group` instance, you also need a `Monoid` instance, because `Monoid` is a superclass of `Group`. And `Monoid` has good examples.

hdgarrood:
Perhaps it's just my lack of imagination, but the only instances I can think of which lead to this problem are ones like `Module Int Int` or `Module Number Number`, or generally `Module R R` for some ring `R`. What about if we say you should generally avoid defining instances of the form `Module R R` for a ring `R`, and instead let people use the `Additive` newtype if they want that instance? I think with that approach, it would be nice that authors of libraries defining types with `Ring` instances then don't need to worry about whether to provide the trivial `Module` instance, and also that instance only needs to be written once.

no-longer-on-githu-b:
@hdgarrood I think you're right. I think many nontrivial use cases don't have multiple good instances. @paf31 What are your thoughts on having a `Group` superclass? This would mean adding `Monoid` and `Group` to purescript-prelude. We can also add group actions.
@paf31 What are your thoughts on providing only left modules?

paf31:
My use case for this is in `purescript-behaviors` where I want to express `integral` in terms of a vector space over the field I'm integrating with respect to (some measure of time). In that case, I would find `Additive` slightly inconvenient to unwrap each time, and `Module r r` is the most common case. However, writing a wrapper function is trivial (I already have one in fact) for the rank-1 case. So I think I prefer to use `Group` as a superclass.

no-longer-on-githu-b:
Then we need to add `Group` and `Monoid`. We can move over just `Data.Monoid` from purescript-monoid, and it shouldn't be a breaking change other than dependency versions. I can also add group actions while I'm at it.

garyb:
We do have the option of not specifically adding it to the prelude, and introducing it as another core library instead. Nothing needs to move then, although we would still need to introduce `Group` somewhere.

hdgarrood:
I'm with @garyb here: I think it would be a bit strange to say that this deserves a spot in the Prelude but not, say, `Maybe`, `Either`, or `Tuple`. I can't really see myself using this all that often except for in a few particular domains.

src/Data/Module.purs
friedbrice: I'd prefer `Module s m` so that `Module s` could be used as a constraint kind.

JordanMartinez:
Sounds like the decision here was to create a repo specifically for this. Would we need to make any breaking changes here to support that? It's been some time since this conversation ended, so I'm not sure what's still relevant and what isn't.

src/Data/Module.purs
```diff
@@ -0,0 +1,62 @@
...
+infixl 6 madd as $+
+infixl 6 msub as $-
+infixl 7 mmul as $*
+
+-- | `mmulFlipped` is `mmul` with the arguments flipped. For example:
+-- |
+-- | ```purescript
+-- | twiceAsLong = (_ *$ 2.0)
+-- | ```
+mmulFlipped :: forall m s. Module m s => m -> s -> m
```
friedbrice, line 41: Switch the order of type parameters, since I believe `Module s` will end up being a more useful partial application.

src/Data/VectorSpace.purs
```diff
@@ -0,0 +1,23 @@
...
+import Data.Module (class Module)
+import Data.Unit (Unit, unit)
+
+-- | The `VectorSpace` class is for modules that support division by a scalar.
+-- |
+-- | Instances must satisfy the following laws in addition to the `Field` and
+-- | `Module` laws:
+-- |
+-- | - Multiplicative inverse: `a /$ (b / c) = a *$ (c / b)`
+class (Field s, Module m s) <= VectorSpace m s | m -> s where
```
friedbrice, line 14: Ditto. Switch order of type parameters.

JordanMartinez:
Should this PR be closed? It sounds like the final decision was to publish this code as a separate PR because it shouldn't be included in `Prelude`. However, the name, [purescript-module](https://pursuit.purescript.org/packages/purescript-modules/5.0.0), is already taken, even if it hasn't been updated to the latest version of `prelude`.

PR #137, opened 2017-08-28
To boolean

Risto-Stevcev:
Adds a `toBoolean` function to convert a HeytingAlgebra to a Boolean value


garyb:
I think if this function were to be provided it should be named something like `toBooleanTrueBiased`... I'm not sure it's a good thing to have though.

`BooleanAlgebra => b -> Boolean` makes a bit more sense, since at least it's not squashing 2-or-more values into 2 like the case with this function... but even still. :man_shrugging:

Risto-Stevcev:
Yeah I wasn't sure about this one TBH. I think I'll just keep this inside the `Data.BooleanEq` module where it makes the most sense

PR #138, opened 2017-08-28
Added xor

Risto-Stevcev:


garyb:
I think this is also only sensible for `BooleanAlgebra`.

hdgarrood:
Do we have a list of motivating examples of types which have `HeytingAlgebra` instances but no `BooleanAlgebra` instances somewhere?

Risto-Stevcev:
A lot of three-valued logics aren't `BooleanAlgebra`s, like the Kleene and Łukasiewicz logics

hdgarrood:
Sorry this has sat for such a long time. I think the function should have a BooleanAlgebra constraint rather than a HeytingAlgebra constraint, and also I don't think this function is important enough to deserve an operator. We also avoid using unicode (neither unicode syntax e.g. for forall, nor unicode operators) in the core libraries. The reason for this is that it's a pain to enter unless you've set your editor up specially, and we don't want to put people through that to be able to contribute. I'd be happy to make those changes and then merge this if another core team member agrees.

I think it's also worth noting that this implementation is likely to be slow compared to, say, using `/=` for `Boolean`, or using the `xor` in `Data.Int.Bits`. We do already have "general-but-slow" operations in a few places already, though, such as `find` for Foldable, so I think that's acceptable.

hdgarrood:
I'm going to close this for now due to lack of interest.

PR #141, opened 2017-10-02
add ap = apply law

Thimoteus:


paf31:
👍 Makes sense to me

garyb:
I know it's exactly the same thing, but wouldn't people usually say `apply = ap` (since you define `apply` for the instance but `ap` is derived)? Either way 💯

Thimoteus:
Ah. I was thinking putting `ap` on the LHS since it's defined in the same module. Something about reading left-to-right, seeing `ap`, asking "wait, what's that? oh, it's a few lines below," then reading `= apply`.

garyb:
Thanks! Forgot about this one until a relevant discussion popped up today :wink:

PR #143, opened 2017-10-25
Added compose math symbol

Risto-Stevcev:


paf31:
We have stuck to non-Unicode in the core libraries so far. That might just be because we always have, but we should make a decision about it, since this would be the first use of it.

michaelficarra:
I don't see any reason to avoid non-ASCII aliases. It's not like you're forcing user code to be written that way.

paf31:
I can think of the following reasons:

- It's harder to read in my opinion.
- It's harder to input.
- It's intimidating for beginners. I remember being put off by Agda's syntax, for example.
- It's another way to do something we can already do, and we should have one preferred way.

We're not forcing code to be written this way, but it will encourage it.

Risto-Stevcev:
Why is this any different than having these though? `∷ ← → ⇐ ⇒ ∀`
All of those concerns apply to those also

Thimoteus:
Are there core libs that use unicode symbols like ∀ and →?

paf31:
No, we don't use them in the core libraries.

hdgarrood:
It looks like we've come to a decision then, so I'll close this.

PR #150, opened 2017-12-14
Make function composition stack safe

safareli:
TODO
- [ ] add more tests
- [ ] add benchmarks
- [ ] add comments to code

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,8 +7,8 @@
   },
   "devDependencies": {
     "eslint": "^3.17.1",
-    "purescript-psa": "^0.5.0-rc.1",
-    "pulp": "^10.0.4",
+    "purescript-psa": "^0.5.1",
+    "pulp": "^11.0.0",
     "rimraf": "^2.6.1"
   }
 }
--- a/src/Control/Semigroupoid.js
+++ b/src/Control/Semigroupoid.js
@@ -1,0 +1,31 @@
+"use strict";
+
+// https://medium.com/@safareli/stack-safe-function-composition-85d61feee37e
+var runComposition = function(composition, x) {
+  var root = composition;
+  var val = x;
+  var stack = [];
+  for (;;) {
+    if (root._0 !== undefined){
+      stack.push(root._1);
+      root = root._0;
+    } else {
+      val = root(val);
+      if (stack.length === 0) {
+        return val;
+      }
+      root = stack.shift();
+    }
+  }
+};
+
+exports.functionCompose = function(f) {
+  return function(g) {
+    var res = function composition(x) {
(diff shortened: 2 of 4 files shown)


safareli:
I got this idea just today and couldn't hold to build and open PR 🗡 

It's sort of composition of work done in Eff http://github.com/purescript/purescript-eff/pull/31 and the post I wrote a while ago about [Stack safe Function composition](https://medium.com/@safareli/stack-safe-function-composition-85d61feee37e).

/cc @natefaubion @paf31

It's worth noting that whenever it's known that composition is happening for function type, compiler is doing optimization so this FFI function will not be called at all, that's why I use `Semigroupoid` constraint in test to force use of the FFI function.

paf31:
Sorry, but in my opinion this is totally unsuitable for Prelude. As I said on the `eff` PR, I think it's interesting, and a worthwhile addition in the form of a separate library, but it's just not the right sort of approach for the core libraries.

hdgarrood:
I'm with @paf31 on this, I think, especially since this only seems to have an effect in rare cases (I don't remember the last time I wrote a function with a `Semigroupoid` constraint). This approach seems risky to me from a maintenance perspective too; I imagine it might be quite easy to accidentally break this in the future if this code path is only rarely being hit.

safareli:
Fair points 🙌

matthewleon:
@safareli did you end up doing anything with this in terms of making a separate lib? I think there could be some really handy uses.

paf31:
I don't think this will work for `Builder`.

safareli:
Nope, how you think this could be used as separate lib?

matthewleon:
Function newtype, perhaps?

matthewleon:
> I don't think this will work for Builder.

`Builder` as it exists now, no... But if someone wrote a separate stack safe Builder lib which used this under the hood, shouldn't that work out?

safareli:
Did you get the stack overflow by using Builder?

matthewleon:
Yes, if you compose enough `Builder`s you will get a stack overflow. The `Category` instance is just `newtype` derived from `Function`.

matthewleon:
(or rather, the `Semigroupoid` instance)

safareli:
Sure, it's less likely to happen, so was curious if you had practical case where it happened. 
I will try to create a lib then

matthewleon:
Don't worry too much about it, as I have another solution for my situation. I just think this particular solution is quite interesting, and probably could be applied in other situations, too.

safareli:
I have published the lib here https://github.com/safareli/purescript-stacksafe-function

PR #151, opened 2017-12-15
Add UInt support

parsonsmatt:
This PR adds instances and support for the `UInt` prim type added in [this compiler PR](https://github.com/purescript/purescript/pull/3169#issuecomment-351932284).

### DIFF
--- a/src/Data/EuclideanRing.purs
+++ b/src/Data/EuclideanRing.purs
@@ -42,7 +42,7 @@ import Data.Semiring (class Semiring, add, mul, one, zero, (*), (+))
 -- | reason not to, `Field` types should normally use this definition of
 -- | `degree`.
 class CommutativeRing a <= EuclideanRing a where
-  degree :: a -> UInt
+  degree :: a -> Int
   div :: a -> a -> a
   mod :: a -> a -> a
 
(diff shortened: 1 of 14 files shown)


src/Data/EuclideanRing.purs
```diff
@@ -42,7 +42,7 @@ import Data.Semiring (class Semiring, add, mul, one, zero, (*), (+))
 -- | reason not to, `Field` types should normally use this definition of
 -- | `degree`.
 class CommutativeRing a <= EuclideanRing a where
-  degree :: a -> Int
+  degree :: a -> UInt
```
parsonsmatt, line 45: Not sure if this is exactly necessary but it does capture the "no negative stuff" rule. An unbounded positive int type may be better.

hdgarrood: Yeah; I'd like to mull this over a bit. Relatedly, if we merge this, are we going to want to change other parts of core libraries to use `UInt` where appropriate, like `Data.Array.length`?

parsonsmatt: I'm not 100% sold on it. I only started doing it because I was trying to write a `EuclidianRing` instance for `UInt` and `degree` was out of range for `Int` :sweat_smile:

PR #162, opened 2018-03-24
ADD: Bounded Number instance.

erisco:
Because `∀x::Number. x ≠ NaN ⇒ x ≤ ∞ ∧ -∞ ≤ x`.

### DIFF
--- a/src/Data/Bounded.js
+++ b/src/Data/Bounded.js
@@ -5,3 +5,6 @@ exports.bottomInt = -2147483648;
 
 exports.topChar = String.fromCharCode(65535);
 exports.bottomChar = String.fromCharCode(0);
+
+exports.topNumber = Number.POSITIVE_INFINITY;
+exports.bottomNumber = Number.NEGATIVE_INFINITY;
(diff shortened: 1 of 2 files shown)


garyb:
This was proposed before, see #116 for previous discussion

erisco:
The proposal there was for `top = Number.MAX_VALUE` and `bottom = Number.MIN_VALUE` which is not what I am proposing here.

paf31:
👍 from me, actually. The linked post says "Number is meant to represent the real numbers", which I think is true in spirit, but (NaN aside), it seems like it'd be useful to change that to the extended reals with +/- infinite, have this instance and be able to use a `Min` or `Max` monoid, for example.

garyb:
Sorry, missed your response to this! Using infinities seems fine to me too since it obeys the laws (ignoring NaN), obviously if someone tries to enumerate `top` to `bottom` they're going to have a bad time, but it's not a `BoundedEnum` so there's no implication that should be a thing anyway.

chexxor:
Thanks for working on this, @erisco! I'd been curious about it for awhile, myself. :)

PR #164, opened 2018-04-13
Add `Monoid` class & newtypes, `Semigroup` newtypes

### LINKED ISSUE #157: Semigroup newtypes
Haskell `base` exports some handy `newtype`s in `Data.Semigroup`. These are especially useful combined with `Foldable1`, see this PR for an example: https://github.com/purescript/purescript-foldable-traversable/pull/93

The useful `newtypes` include `Min`, `Max`, `First` and `Last`.

I'd be happy to make a PR here for these, and then amend my PR on `Foldable1`, above, to use them.

### LINKED ISSUE #163: Add `Monoid` to prelude?
Perhaps we should just deprecate the library and roll it into `Prelude`, since `Monoid` is super common?

I'd also be in favour of adding @matthewleon's `Semigroup` newtypes in then too, as it seems sensible to follow the same pattern we use for monoids, assuming we do move those over at the same time.

/cc @paf31 (if you're interested at all), @kRITZCREEK, @LiamGoodacre + anyone else with thoughts on it

garyb:
Resolves #163, resolves #157

I generalized the `Endo` monoid to any category while I was at it - I know we have `Join` that does the same thing, but we can probably just do away with that instead, since that's tucked away in `-profunctor`.

This will fail to build just now as it uses `Eq1` and `Ord1` deriving that is only present in 0.12.

### DIFF
--- a/src/Data/Monoid/Additive.purs
+++ b/src/Data/Monoid/Additive.purs
@@ -1,0 +1,44 @@
+module Data.Monoid.Additive where
+
+import Prelude
+
+import Data.Eq (class Eq1)
+import Data.Ord (class Ord1)
+
+-- | Monoid and semigroup for semirings under addition.
+-- |
+-- | ``` purescript
+-- | Additive x <> Additive y == Additive (x + y)
+-- | mempty :: Additive _ == Additive zero
+-- | ```
+newtype Additive a = Additive a
+
+derive newtype instance eqAdditive :: Eq a => Eq (Additive a)
+derive instance eq1Additive :: Eq1 Additive
+
+derive newtype instance ordAdditive :: Ord a => Ord (Additive a)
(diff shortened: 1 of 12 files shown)


src/Data/Monoid/Additive.purs
```diff
@@ -0,0 +1,44 @@
+module Data.Monoid.Additive where
+
+import Prelude
```
LiamGoodacre, line 3: Shouldn't import itself :smile_cat: (similarly with the others files)

garyb: These ones are ok, as the `Prelude` module doesn't re-export them :)

LiamGoodacre: Oh yes, of course!

LiamGoodacre (approved):

PR #166, opened 2018-04-13
Update build dependencies

garyb:
Need to figure out what arguments are necessary for publishing before merging this, since there's manifest/resolutions arguments now.


LiamGoodacre:
I think it's just license and repository in bower.json, which we have already.

garyb:
Yeah, I think that's it too, I just want to test something locally on a library that needs publishing to be sure I have it in a form that works :)

garyb:
I don't need to do anything here - `pulp publish` provides the appropriate arguments in the newer versions anyway.  :smile:

PR #167, opened 2018-04-13
Class law changes

### LINKED ISSUE #132: Clean up numeric hierarchy
After #128, we now have a couple of small problems:

* `Field` is exactly `CommutativeRing` and `DivisionRing`; forcing people to declare an extra `Field` instance doesn't really add anything.
* `Field` should have `DivisionRing` as a subclass, since every `Field` is a `DivisionRing`.

As far as I can tell, there are two options for addressing this:

**Option 1:** Change `Field` as follows:

```purescript
```

So that no additional `Field` instances should be declared anywhere, and instead the single `Field` instance allows people to use a `Field` constraint wherever a type has both `DivisionRing` and `CommutativeRing` instances. If we did this, it would only be a breaking change for people who had declared `Field` instances for types in their own code; functions with `Field` constraints should continue to work as before.

**Option 2:** Wait for constraint kinds and define `Field` in terms of a class `class Both (a :: Constraint) (b :: Constraint)` (is that accurate?)

### LINKED ISSUE #153: Recommend extensional equality for `Eq` instances
While I think requiring an `Eq` instance to define an equivalence relation is definitely correct, I've felt for quite a long time that the bar for an `Eq` instance to be considered valid should actually be a little higher, based on e.g. the way `Eq` gets used generically. For example, currently, if we fix any nonzero `n :: Int`, the following would be a perfectly good `Eq Int` instance:

```purescript
instance Eq Int where
  eq = unsafeRefEq `on` (_ `rem` n)
```
(where `rem` is like `mod` except the result of ``x `rem` n`` is always between `0` and `n`, i.e. if the result is nonzero, then its sign matches the sign of the second argument.)

That would mean that the behaviour of of `nub [1,3,5,7,8,9]` (where `nub` is using the `Eq` constraint) could be any of `[1]`, `[1,8]`, `[1,3,5]`, `[1,3,8]`, `[1,3,5,7,8]`, or `[1,3,5,7,8,9]` for different choices of `n` above.

I think what I'm about to propose is a bit too hand-wavy to be described as a 'law', but it is a guideline I've been following for a while, and I think it makes a lot of sense and also matches up well with how `Eq` is actually used in practice. How do people feel about adding something like the following to the `Eq` docs:

> While not a strict requirement, when defining an `Eq` instance for some type `T`, it is recommended that for any pair of values `x :: T` and `y :: T`, the expression `x == y` should evaluate to `true` if and only if `x` and `y` are indistinguishable via the rest of the functions and instances which form the public API of `T`. In particular, if there is an (exported) function `f :: T -> U` for some other type `U` which itself has an `Eq` instance, then whenever `x == y` we should have `f x == f y` as well.
>
> This allows us to define `Eq` instances for data structures which hide their representation and may have more than one possible internal representation for a given value, such as `Map`, while still being sufficiently restrictive to ensure that `Eq` instances mostly act in an intuitive manner.

### LINKED ISSUE #161: Change mod for Int
For example, -3 mod 9 should be 6 (since 6 - 9 = -3) but in PS we have `mod (-3) 9 == (-3)`, which ultimately stems from JS's `-3 % 9 === -3`

Some people suggest using the following FFI:

```javascript
function mod(n, m) = {
  return ((n % m) + m) % m;
}
```

References:
http://www.seqmedia.com/2011/03/16/javascript-negative-numbers-modulus-bug/
https://stackoverflow.com/questions/4467539/javascript-modulo-gives-a-negative-result-for-negative-numbers

garyb:
Resolves #153, resolves #132, resolves #161

@hdgarrood perhaps you could take a look at this one?


src/Data/Eq.purs
```diff
@@ -18,6 +18,9 @@ import Data.Void (Void)
 -- | class due to the presence of `NaN`, since `NaN /= NaN`. Additionally,
 -- | computing with `Number` can result in a loss of precision, so sometimes
 -- | values that should be equivalent are not.
+-- |
+-- | It is generally recommended to declare `Eq` instances where
+-- | `x == y && f x == f y` but not a strict requirement.
```
LiamGoodacre: Should this `&&` be `implies`?

Also perhaps this should be more explicit that the two `==` are potentially using different instances.

Is it worth giving a short example where this isn't true? Like:
```
newtype Identical a = Identical a
unidentical (Identical a) = a
instance eqIdentical :: Eq (Identical a) where eq _ _ = true
```
This `Eq` instance isn't isomorphic to the `Eq` of the underlying type.  Therefore, `unidentical` is a function that lets you recover the distinction between values.  If `f ~ unidentical` then ```x == y `implies` f x == f y``` is false.

garyb: Ah yeah, `implies`, or just `(x == y) == (f x == f y)` I guess.

That example `Identical` example is perfectly correct, but perhaps we could come up with one that seems more "useful" :smile:. I'll have a think, as it still needs to be trivial enough that it requires not much more than the definitions here.

LiamGoodacre: Well, in general `f x == f y` doesn't need to imply `x == y`.  That would be a requirement of injectivity for `f`?

LiamGoodacre: Unless you're saying "for all injective `f`, the following should be true...".

garyb: Yeah, right again! `f = const unit` is going to violate most instances of that. I guess "injective `f`" can be included in the description.

LiamGoodacre: And then we'd possibly need to say what kind of injectivity we're talking about, because if it's with respect to `Eq`, then `Identical`/`unidentical` aren't valid `f`s -  because the `Eq` isn't iso to the inner type's `Eq`.  That is, the equality semantics need to be talking "in the same language" to make sense.
If it's injective with respect to definitional equality, then those functions are injective. But that messes things up 😅.

Perhaps a less wordy way to say all this is:
> An `Eq t` instance should ideally be consistent with the `Eq` instances of everything that can be injected into and projected out of `t`, should those instances exist.

LiamGoodacre: Kmett said it better: "structural and reflexive" 😸 
https://www.reddit.com/r/haskell/comments/87mf4v/is_it_considered_acceptable_for_eq_instances_to/

hdgarrood:
I think it might be best to leave the first line as is, especially since we are adding the recommendation that `x == y` should imply `f x == f y`. If the class were intended just for any equivalence relation, then we wouldn’t be adding that recommendation. Also saying it’s for ‘types which

hdgarrood:
Oops.

....‘for types which support an equivalence relation’ doesn’t really tell you anything - since `const (const true)` is an equivalence relation which could be used for any type.

garyb:
@hdgarrood I just typed a bunch of stuff but then actually read what you wrote properly :wink:

Reverting the change to the first part is fine by me. I changed it because the laws we require are just those of an equivalence relation, but you're right that with the extra recommendation that is a bit strange, so :+1:.

garyb:
I've done the implementation changes for `Int` mod/div using the E-definition, but could use a hand with revising the laws accordingly (I didn't have much luck finding that paper you mentioned in #161 with the definitions in @hdgarrood).

hdgarrood:
The laws should stay the same, it’s `div` which needs to change - I think the current `div` is the T-definition. I think ideally the tests should check all 4 combinations of positive and negative dividends and divisors.

hdgarrood (commented):
I’ve just realised this `mod` is incorrect. We should have ``a `mod` (-b) == a `mod` b``, for the E-definition, I think. I’ll have a think about how best to implement the E-definition of `div` and `mod`.

I think we should also provide `quot :: Int -> Int -> Int` and `rem :: Int -> Int -> Int` which would be the same as the current `div` and `mod` and ideally would have inlining rules, in order to assuage people who are worried about perf, and/or for ease of porting code originally written in other languages.

hdgarrood:
Also, I think having the recommendation say that `x == y` should imply `f x == f y` would be closer to what we want. Consider `Map`, for instance: the `Eq (Map k v)` instance definitely isn't structural, but it is the 'correct' instance for that type, and it does follow the rule that `x == y` implies `f x == f y` for pretty much all functions `f` exported by that module.

garyb:
I revised the `Eq` description again to use something Liam mentioned, it avoids the structural mention and instead phrases it in terms of things that can be injected/projected from the type, which I think is still sensible interpretation of what happens for `Map`. Plus I like that it's a bit more vague :wink: given that it's a guideline rather than a law, and ``(x == y) `implies` (f x == f y)`` looks more law-y.

garyb:
Any thoughts on the current description @hdgarrood / @LiamGoodacre? I could just revert the change to `Eq` and leave it as it was as another alternative :smile:

LiamGoodacre:
Yeah let's revert the Eq description and get the rest of the changes in.  We can continue the discussion of Eq later if needed.

garyb:
Ok cool. Closing this one as it's a bit of a mess! (Since we did `Int` changes elsewhere)

PR #168, opened 2018-04-15
Switch to Euclidean division for Int, resolves #161

### LINKED ISSUE #161: Change mod for Int

hdgarrood:
Also provide `quot` and `rem`, like Haskell does, for users who do want
truncating division - the one which matches what JS does.

I've temporarily exported `intDiv` and `intMod` so that I can use those
in the tests and the compiler won't 'inline' different definitions of
them; we'll want to modify the compiler to change this before merging.

I've also done a more detailed writeup [in a separate repo](https://github.com/hdgarrood/integer-divisions/blob/c4d907a109a83df5b61966e607e6cd068b22c6b3/main.js) but I thought that might be a bit much for these docs.


src/Data/EuclideanRing.js
```diff
@@ -4,14 +4,29 @@ exports.intDegree = function (x) {
   return Math.min(Math.abs(x), 2147483647);
 };
 
+// See the Euclidean definition in
+// https://en.m.wikipedia.org/wiki/Modulo_operation.
 exports.intDiv = function (x) {
+  return function (y) {
+    return Math.sign(y) * Math.floor(x / Math.abs(y));
```
garyb: `Math.sign` is ES2015 unfortunately

src/Data/EuclideanRing.purs
```diff
@@ -77,3 +100,38 @@ lcm a b =
...
+-- | mod 2 3 == 2
+-- | rem 2 3 == 2
+-- |
+-- | mod (-2) 3 == 1
+-- | rem (-2) 3 == (-2)
+-- |
+-- | mod 2 (-3) == 2
+-- | rem 2 (-3) == 2
+-- | ```
+foreign import rem :: Int -> Int -> Int
```
garyb, line 137: Perhaps we should provide `quot` and `rem` in `purescript-integers`, since they are `Int` specialised, rather than constrained to `EuclideanRing`?

src/Data/EuclideanRing.purs
```diff
@@ -41,6 +45,25 @@ import Data.Semiring (class Semiring, add, mul, one, zero, (*), (+))
...
+-- | negative dividends or divisors. The most common definitions are "truncating"
+-- | division, where the result of `a / b` is rounded towards 0, and "Knuthian"
+-- | or "flooring" division, where the result of `a / b` is rounded towards
+-- | negative infinity. A slightly less common, but arguably more useful, option
+-- | is "Euclidean" division, which is defined so as to ensure that ``a `mod` b``
+-- | is always nonnegative. With Euclidean division, `a / b` rounds towards
+-- | negative infinity if the divisor is positive, and towards positive infinity
+-- | if the divisor is negative. Note that all three definitions are identical if
+-- | we restrict our attention to nonnegative dividends and divisors.
+
```
garyb: Need a `-- |` here I think :smile:

garyb:
This looks good to me, aside from those few comments above. I've opened a PR to remove `mod` inlining for `Int` already: https://github.com/purescript/purescript/pull/3309 and it appears we didn't inline integer division already... I found that a bit of a surprise!

src/Data/EuclideanRing.js
garyb: I guess the simplest alternative is something like:
``` javascript 
return y > 0 ? Math.floor(x / y) : -Math.floor(x / -y);
```

garyb:
Ah no, we do, but it's done differently than the mod case. I'll update that PR accordingly.

src/Data/EuclideanRing.js
hdgarrood: Ah, thanks, I wasn't aware of that.

src/Data/EuclideanRing.purs
hdgarrood: Yeah, that wouldn't be a bad option. I think it makes sense to put them here, because they do form a valid `EuclideanRing` instance (even if we're no longer using it as the default instance for `Int`), and also because these functions were available from Prelude before, but I don't have a strong preference.

garyb:
Thanks again for working on this!

hdgarrood:
Happy to :) just one thing though - I just realised that the code here still exports `intDiv` and `intMod`, and I didn't mean for that to stay in the final version. I guess now that the compiler PR to stop inlining `div` and `mod` has been merged, we should remove those exports and update the tests to use `div` and `mod` instead.

garyb:
Yeah, thanks for the reminder - I went ahead and did that already. I did move `quot` and `rem` to `Data.Int` too, since I felt it was a little strange to have some monomorphic functions in a module that is otherwise purely polymorphic. I'll include a note in the release notes when we finally get around to it pointing out the breaking change and the option to use those functions.

PR #170, opened 2018-04-30
Update `Field`

### LINKED ISSUE #128: Re-add DivisionRing?

### LINKED ISSUE #132: Clean up numeric hierarchy

garyb:
Resolves #132

### DIFF
--- a/src/Data/Field.purs
+++ b/src/Data/Field.purs
@@ -15,14 +15,9 @@ import Data.Semiring (class Semiring, add, mul, one, zero, (*), (+))
 
 -- | The `Field` class is for types that are (commutative) fields.
 -- |
--- | Instances must satisfy the following law in addition to the
--- | `EuclideanRing` laws:
--- |
--- | - Non-zero multiplicative inverse: ``a `mod` b = zero`` for all `a` and `b`
--- |
--- | If a type has a `Field` instance, it should also have a `DivisionRing`
--- | instance. In a future release, `DivisionRing` may become a superclass of
--- | `Field`.
-class EuclideanRing a <= Field a
+-- | `Field`s are exactly `EuclideanRing` + `CommutativeRing` so this class
+-- | exists as a convenience, so a single constraint can be used when field-like
+-- | behaviour is expected.
+class (EuclideanRing a, CommutativeRing a) <= Field a
 
-instance fieldNumber :: Field Number
+instance fieldNumber :: (EuclideanRing a, CommutativeRing a) => Field a


hdgarrood:
Every field is a euclidean ring, so I think it might be best to keep the EuclideanRing constraint. In fact with what's in this PR, you wouldn't be able to write a function which uses `/` and which only has a `Field` constraint, right?

How about this instead:

```purescript
class (EuclideanRing a, DivisionRing a) <= Field a
instance (EuclideanRing a, DivisionRing a) => Field a
```

hdgarrood:
Also this relates to #132, not #128, right?

garyb:
I did wonder about that, in the other PR I had it with the `EuclideanRing` constraint but then looking at it again today I thought that might have been a mistake :smile:

hdgarrood:
Also I think it might be worth mentioning that every type with law-abiding `DivisionRing` and `CommutativeRing` instances only really has one sensible law-abiding `EuclideanRing` implementation, where `div a b = a * recip b`, `mod _ _ = zero`, and `degree _ = 1`; see <https://a-guide-to-the-purescript-numeric-hierarchy.readthedocs.io/en/latest/euclidean-rings.html>, where it says "start from "Let F be a field, and suppose f:F∖{0}→N is a euclidean function". `degree` must be constant; any return value will work, so you might as well choose 1.

hdgarrood:
Sorry, I didn't notice you had updated this, but there's a couple of things I think are not quite right here (in the docs). Do you mind if I send another PR?

hdgarrood:
Oh actually, `EuclideanRing` implies `CommutativeRing`, so adding the `CommutativeRing` superclass won't do anything. With this commit, we have a `Field Int` instance, which is definitely wrong. It should have `EuclideanRing` and `DivisionRing` superclasses. I'll send a PR in a moment.

garyb:
Oops, glad someone is paying attention :wink:

PR #171, opened 2018-04-30
Instances for Record

i-am-tom:
Hullo!

This PR adds what I think/hope are reasonable implementations for all the standard type classes applicable to records. Please, please, please tear this apart, in case I've missed anything - being in the prelude, it is very tricky to test, and I'm worried I'll have missed something the type-checker wouldn't have caught.

However, if you can get the tests running, you probably don't need to worry. I, personally, couldn't, but I might be missing something obvious?

Thanks!


MonoidMusician:
Hey looking nice, thanks for your work!!

For things like `HeytingAlgebra => BooleanAlgebra`, I think you actually will need to make sure that each type in the row also implements `BooleanAlgebra` in order for the whole record to satisfy the laws (same thing for `Ring => CommutativeRing`).

So this might mean you'll have to implement `EachIsBooleanAlgebra` that checks that for each type in the row...

... or the other option is you parameterize `HeytingAlgebraRow` over a argument that specifies whether you want `BooleanAlgebra` from each type or just `HeytingAlgebra`. (In general I like the idea of encoding laws as parameters on the typeclass, but that's obviously out of scope here haha.)

P.S. Especially if you go with the former method, the you could just use `HeytingAlgebra (Record row)` as the constraint instead of `RL.RowToList row list, HeytingAlgebraRow list row row focus`. Not a big deal though :)

Also, I'm curious why `focus` is exposed as a parameter to each class?

src/Data/HeytingAlgebra.purs
```diff
@@ -1,9 +1,31 @@
...
+  , (&&)
+  , (||)
+
+  , class HeytingAlgebraRow
+  , ffRecordImpl
+  , ttRecordImpl
+  , impliesRecordImpl
+  , conjRecordImpl
+  , disjRecordImpl
+  , notRecordImpl
```
garyb: Maybe we could drop `Impl` from all these kinds of names - usually that's for FFI things where there's a non-`Impl` version in scope too.

Perhaps the classes should be using `Record` in their name rather than `Row` too, since they are specialised to records in particular.

garyb: Thinking about it, maybe all these should be defined in modules like `Data.HeytingAlgebra.Record` or something like that, that way they don't have to be exported from these "primary" modules, given that there's not really a need to use these functions outside of the record instances.

src/Data/Internal/Record.purs
```diff
@@ -0,0 +1,14 @@
+module Data.Internal.Record where
```
garyb: `Data.Internal` is a bit of a weird namespace - perhaps it could be `Unsafe.Record` like we have `Unsafe.Coerce` and `Unsafe.Reference` (`unsafeRefEq`).

src/Data/Internal/Record.purs
```diff
@@ -0,0 +1,14 @@
...
+
+-- | *Really* unsafely get a value from a record. You really shouldn't be using
+-- | this function unless you know what you're doing.
```
garyb: Perhaps we could refine these comments a bit :smile:. The `unsafeCoerce` comment is pretty detailed actually: https://pursuit.purescript.org/packages/purescript-unsafe-coerce/3.0.0/docs/Unsafe.Coerce#v:unsafeCoerce

At least talking about the purpose / how to use them safely (such as it is) would be good.

src/Data/Internal/Record.purs
i-am-tom: @LiamGoodacre mentioned potentially just merging `Data.Record.Unsafe` from `purescript-record` in here?

garyb: Sure, that works.

i-am-tom: which I guess also clears up the comment below, too :D Will sort this out (and remove `focus` from all the classes - thanks @MonoidMusician) at lunch :+1:

src/Data/HeytingAlgebra.purs
i-am-tom: Only problem here is that they all require things to have `HeytingAlgebra` constraints, which would cause a circular dependency :(

hdgarrood: In that case, how about putting all of this in `Data.HeytingAlgebra.Internal`, and then re-exporting the important bits from `Data.HeytingAlgebra`?

garyb:
Did some more work on this:

- didn't end up adding the `Internal` modules because that caused issues with being able to derive `Eq` (since it was no longer `Data.Eq.Eq`)
- removed `focus` parameters
- renamed things from `Row` to `Record` as appropriate
- added new record classes for `BooleanAlgebra` and `CommutativeRing`, since the way they were done in here upgraded any record-of-`HeytingAlgebra` to be a record-of-`BooleanAlgebra` regardless of whether the values had `BooleanAlgebra` instances (and same for `CommutativeRing`)
- Moved over `Record.Unsafe`'s function

garyb:
I'll merge this for now and we can make any further changes or fixes that are needed on the branch. Thanks @i-am-tom!

i-am-tom:
You're brilliant, @garyb - thanks so much!

PR #172, opened 2018-05-01
Explain the magic numbers for Bounded Int.

chexxor:
This question came up in the chat room.

### DIFF
--- a/src/Data/Bounded.js
+++ b/src/Data/Bounded.js
@@ -1,5 +1,7 @@
 "use strict";
 
+// Can't use `Number.MIN_SAFE_INTEGER` or `Number.MAX_SAFE_INTEGER` because
+//   those apply to double precision floating point, not 32-bit signed integers.
 exports.topInt = 2147483647;
 exports.bottomInt = -2147483648;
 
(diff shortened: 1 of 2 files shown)


hdgarrood:
I think it would make more sense to put this on the `Bounded Int` instance in the `purs` file, so that it shows up in the HTML docs. Also, I don't think we should refer to floating point at all, since `Int` really has very little to do with floating point; the fact that `Int` is sort of like `double` underneath is only an implementation detail for the JS backend.

Here's what I'd suggest instead:

> The `Bounded Int` instance has `top :: Int` equal to `2^31 - 1`, and `bottom :: Int` equal to `-2^31`, since these are the largest and smallest integers representable by twos-complement 32-bit integers respectively.

chexxor:
Great ideas, @hdgarrood - I've moved the docs and taken your suggestion. Your explanation is more succinct and the reason is more satisfying to the curious mind coming to find the answer.

Hope it's possible to squash these into one commit - would be weird to add 3 commits to implement this patch, haha.

hdgarrood:
Thanks! And also good idea to clarify this in the docs in the first place. I'd just like to remove the indentation on the second and third lines of your doc-comment, just in case it messes up the formatting in the HTML docs.

We can squash and merge in one go with the merge button now, incidentally, which is nice.

hdgarrood:
Oh also, I think it might also be good to preserve the backticks.

garyb:
Thanks!

PR #177, opened 2018-06-06
Optimization for Eq Array

jazmit:
Running Eq on large nested data structures can be expensive (think rose trees with 10^5 items), so checking for reference equality before traversing the array can result in some significant gains.


LiamGoodacre (approved):

garyb:
Thanks!

Issue #176, opened 2018-05-30
Enable Ord Record instance and clarify its equivalence to compiler derived Ord


natefaubion:
Compiler derived Ord for records is the same as RowToList ordering. I'd suggest we enable it (it's currently commented out) and note the equivalence between:

```
newtype MyRecord = MyRecord {...}
```
```
derive instance ordMyRecord :: Ord MyRecord
```
```
instance ordMyRecord :: Ord MyRecord where
  compare (MyRecord r1) (MyRecord r2) = compare r1 r2
```

garyb:
I’d also be fine with this - I wasn’t really sure why the alphabetical label ordering was thought to be a problem, given that’s how derived instances work, and how generics did.

It needs documenting somewhere of course, but I don’t think it’s particularly strange - it seems “obvious“  that there would need to be some canonical ordering, given record labels are a set.

PR #180, opened 2018-07-16
Add/finish Ord instance for Records

LiamGoodacre:
Fixes https://github.com/purescript/purescript-prelude/issues/176

It's very annoying that we didn't make this instance, especially when it disrupts upgrading in some cases: [removal-of-deriving-for-records-in-generics-rep#3](https://purescript-users.ml/t/removal-of-deriving-for-records-in-generics-rep/197/3).

### DIFF
--- a/src/Data/Ord.purs
+++ b/src/Data/Ord.purs
@@ -12,14 +12,20 @@ module Data.Ord
   , abs
   , signum
   , module Data.Ordering
+  , class OrdRecord, compareRecord
   ) where
 
-import Data.Eq (class Eq, class Eq1)
+import Data.Eq (class Eq, class Eq1, class EqRecord, (/=))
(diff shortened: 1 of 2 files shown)


garyb:
I'm for it 👍 

Anyone feel strongly against it for any reason? Speak now or forever hold your peas.

hdgarrood:
I’m slightly uneasy about the potential for people to use a type synonym for a particular record type and expect its Ord instance to use the order written in the type synonym declaration, as was discussed somewhere a while ago (I can’t be bothered to hunt for it right now). This is a fairly minor objection though, and instance is clearly useful so :+1:

natefaubion:
@hdgarrood I don't think that's any different than people using `Ord` deriving and expecting the compiler to use the order of the fields in the data declaration. Edit: To clarify, for records.

PR #186, opened 2018-09-18
Update Semigroup.Last's description

tippenein:


LiamGoodacre:
Thanks, but this is the same as https://github.com/purescript/purescript-prelude/pull/184

Issue #188, opened 2018-10-07
Add Hashable typeclass


fehrenbach:
I would like to add a `Hashable` typeclass to the prelude and deriving support for it to the compiler. It is intended as an approximation of equality for use in hash-based algorithms and data structures, such as hash maps, not cryptographically secure digests or password hashes.
```PureScript
class Eq a <= Hashable a where
  hash :: a -> Int
```
There are two reasons for adding it to the prelude instead of having a standalone library:
1. __Dependencies:__ To be useful in general data structures, as many datatypes as possible should be hashable. Ideally, whenever a datatype is `Eq`, it should also have `Ord` and `Hashable` instances, if possible. Most packages depend on the prelude already.<sup>[citation needed]</sup> This should make it easier to convince authors to add `Hashable` instances, because they do not need to add another dependency, all that is required is one more deriving line below the lines for `Eq` and `Ord`.
2. __Performance:__ Usually, the main goal of hashing is to avoid expensive comparisons. One could write a generic hash function in terms of `purescript-generics-rep`. This would be fine for many applications, but for hash-based data structures etc., the hash function should really be as fast as possible.

I am more than happy to make an initial pull request (based on https://github.com/fehrenbach/purescript-unordered-collections/blob/master/src/Data/Hashable.purs) and work on deriving support.

What do people think?

kritzcreek:
I'd definitely like to have this :+1:

fehrenbach:
Great.

I think the only major design decision is whether to `newtype Hash = Hash Int` or not.

garyb:
I'd like the `newtype` personally, as the name at least gives a bit of a semantic clue to the value's intention.

kritzcreek:
Since this is primarily done for performance reasons I'd like to make sure the newtype doesn't incur any annoying penalties, obviously newtype means the underlying representation is an Int, but if we find ourselves mapping the newtype constructor a lot we'll end up with `identity` calls that the compiler currently doesn't optimize.

fehrenbach:
Also, equality comparisons on Int are special-cased by the compiler to emit ===. This is not currently done for (newtype) derived instances for newtypes over Int.
```PureScript
newtype Hash = Hash Int
derive newtype {- or not -} instance eqHash :: Eq Hash

approxEq :: Hash -> Hash -> Boolean
approxEq a b = a == b -- ... return Data_Eq.eq(eqHash)(a)(b);

-- can be avoided by being careful
[...]

PR #189, opened 2018-10-11
Hashable type class

fehrenbach:
See #188.

### DIFF
--- a/src/Data/Hashable.purs
+++ b/src/Data/Hashable.purs
@@ -37,8 +37,7 @@ class Eq a <= Hashable a where
   hash :: a -> Int
 
 instance hashableBoolean :: Hashable Boolean where
-  hash false = 0
-  hash true = 1
+  hash b = if b then 1 else 0
 
 instance hashableInt :: Hashable Int where
   hash n = n
--- a/src/Data/Hashable.purs
+++ b/src/Data/Hashable.purs
@@ -81,7 +81,8 @@ instance hashableRecordCons ::
   , Row.Cons l vt whatev r
   ) => HashableRecord (Cons l vt tl) r where
   hashRecord rlp record =
-    hash field * 31 + hashRecord (RLProxy :: RLProxy tl) record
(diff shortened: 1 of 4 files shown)


kritzcreek:
This looks like a great start! I'm assuming you didn't just come up with these hashing functions and took them from somewhere else? (The Haskell implementation? Some paper?)

I think it would be good to have a few comments referencing where these hashing functions are coming from, to have some form of rationale for them.

fehrenbach:
The one for Number is used by immutable.js, among others.
Haskell uses identity for Int and fromEnum for "small" types.
The String and Array algorithm here is the same as Java, I think.

I can add comments. It'd be even better if someone did some actual research on the matter. We are in the fortunate position that Eq is homogeneous and there is no subtyping, so changing the algorithm is not a breaking change and can be done for any type individually.

fehrenbach:
This now uses the indexed newtype as discussed in #188.

src/Data/Hashable.purs
```diff
@@ -0,0 +1,118 @@
...
+-- | ```PureScript
+-- | (a == b) `implies` (hash a == hash b)
+-- | ```
+-- |
+-- | That is, unequal hash values are a safe approximation of
+-- | inequality. In other words, two objects whose hash values differ,
+-- | are never equal. The reverse is not necessarily true.
+-- |
+-- | Hash values produced by `hash` must not be relied upon to be
+-- | stable accross multiple executions of a program and should not be
```
LiamGoodacre: Typo: `accross` -> `across`

fehrenbach: Thanks!

fehrenbach:
Thanks to @rightfold for pointing out the array hash problem.

JordanMartinez:
@fehrenbach Should this PR be closed in favor of #198?

fehrenbach:
Yeah, probably. I should get back to this at some point...

PR #190, opened 2018-10-18
Add Equiv monoid

Thimoteus:


hdgarrood:
I'm tempted to say that this should go into an external library.

Issue #185, opened 2018-08-23
left-associative `apply` operator


acple:
The `apply` is known as operator `$` that is right-associative.
I am now learning F#, has the 'backward-pipe operator' `<|` with type `(a -> b) -> a -> b` is left-associative.
It is very useful in several scenes.

```PureScript
module Data.Function where
...

infixr 0 apply as $
infixl 1 apply as |$ -- left-associative version, need to consider which symbol use
```

```PureScript
let x = f |$ g a b |$ h c |$ d

-- is equivalent to:
let x = f (g a b) (h c) d

\_ -> myfunc
  |$ parameter aaaa
  |$ expression <$> is
  |$ verylong
```

This is more clear for multi param function application, in limited usecases.

How do you think so?

kedashoe:
like https://pursuit.purescript.org/packages/purescript-prelude/4.1.0/docs/Data.Function#v:(#) ?

natefaubion:
(#) is left associative but the arguments are flipped. A left associative version of `$` lets you group arguments with a single delimiter. That is `foo |$ a + 1 |$ b + 2 |$ c + 3` is equivalent to `foo (a + 1) (b + 2) (c + 3)`.

acple:
these are three apply operators difference:

```PureScript
f a $ g b $ h c === f a (g b (h c)) -- f :: a -> gbhc -> x

f a # g b # h c === h c (g b (f a)) -- f :: a -> y

f a |$ g b |$ h c === (f a) (g b) (h c) -- f :: a -> gb -> hc -> z
```

The third is now don't exist but i occasionally want it!

acple:
Closing this, see #193.

PR #193, opened 2019-01-18
Add left-associative `apply` operator

acple:
see #185, I checked no symbol conflicts in all current psc-package libraries.


chexxor:
Do you have some examples of when you like to use this?

klntsky:
It can be used to avoid parentheses in some cases. [Similar thing in haskell](https://hackage.haskell.org/package/hmatrix-0.16.0.6/candidate/docs/src/Data-Packed-Foreign.html#app). 

Cite:

```
Only useful since it is left associated with a precedence of 1, unlike $, which is right associative. e.g.

someFunction
    `appMatrixLen` m
    `appVectorLen` v
    `app` other
    `app` arguments
    `app` go here

One could also write:

(someFunction 
    `appMatrixLen` m
    `appVectorLen` v) 
    other 
    arguments 
    (go here)
```

acple:
When call multi-param function with long long argument expression, parenthesized linebreak is bit confusing.
it sometimes happen in defining ffi.

```purescript
foreign import example :: ...... (Either a b -> Effect Unit) -> (c -> Effect Unit) -> d -> Effect Unit

now = example (either (failHandler <<< showError) (pure <<< convert <<< readValue)) (myFunc x y) ("aaaaaaa" <> value) -- where I break line?

pr = example
  |$ either
        (failHandler <<< showError)
        (pure <<< convert <<< readValue)
  |$ myFunc x y
  |$ "aaaaaaa" <> value
```
linebreaked args is bit complicated but binaryop solves it, and, I want to avoid using parentheses if possible.

hdgarrood:
Sorry for leaving this for a while. I'm leaning towards saying that this shouldn't be in Prelude, because arguably we have enough different ways of writing function application already: parentheses, `$`, and `#`. It's easy enough to define this operator locally to your project, I think. I'd probably write that example as

```purescript
now = example
  (either
    (failHandler <<< showError)
    (pure <<< convert <<< readValue))
  (myFunc x y)
  ("aaaaaaa" <> value)
```

acple:
Yeah I know that can be easy to define locally, but it is absolutely common operator and I think that makes more readable in certain situations.
`$` and `#` are very useful operator but all can be replaced with nested parentheses at the expense of readability.
Why `$` and `#` are fair but proposed isn't?

natefaubion:
Note, you can use `do` layout for this.

```purescript
pr = example
  do either
        (failHandler <<< showError)
        (pure <<< convert <<< readValue)
  do myFunc x y
  do "aaaaaaa" <> value
```

acple:
wow...I've never thought about use `do` without monads, I think its very hacky but reasonable for argument separation. Is this well-known way in PureScript?

natefaubion:
`do` syntax in PS doesn't necessarily have anything to do with Monads (especially with qualified do). I personally use `do/let` all the time as a less ugly alternative to `let/in`.

As for the operators, the compiler has a weak optimizer, and so we need a named function to inline it (which we have with `apply`). The infix alias is just sugar though and can be defined anywhere. `$` and `#` are both extremely common idioms (across many languages) because you can chain pipelines of function calls together into a single expression. You can't do that with a left associative operator, so it's only an argument delimiter for a single call. I personally have never really wanted or needed this, but that's OK. That's why defining and importing your own Prelude is a much more common thing in PureScript. I've had one in every professional PS project I've been a part of.

hdgarrood:
> Why $ and # are fair but proposed isn't?

This might not be a very satisfying answer, but it's essentially because they've been there for a really long time and |$ hasn't. I think we should start being very conservative with changes to Prelude these days: it's basically at the bottom of the ecosystem dependency chain, so changes can affect the entire ecosystem. There may not be people using it as an operator now as far as we are aware, but it could be rather annoying if some downstream library or some app was already using it for something else, or wanted to start using it for something else.

I'm also a little concerned from a teaching perspective: I think beginners will want to go through Prelude and understand everything in it and know when it's useful (and I think this is reasonable, and something that the design of Prelude should take into account). Putting more stuff into Prelude makes that harder. Also, I think putting something in the Prelude kind of leads to a sort of official endorsement of it: there's an implicit "it's a good idea to use this in your code" attached. In this case, that's something I'd prefer not to do. Increasing the number of different ways of writing any given collection of function applications risks making code harder to read: there's more context you have to develop to make sense of things. This is again likely to cause more difficulty for learners especially.

I should add that this decision is definitely not just up to me, and I can't speak for the other maintainers. I'm just laying out an argument as to why I'd rather not merge this.

acple:
Fair enough.
I knew this is not really necessary. Thank you for the clear reason.
Closing this.

and @natefaubion's suggestion is surprised me. I use `do/let` just from today. Thanks!

PR #197, opened 2019-03-03
Add laws

garyb:
It's been a while since we worked on this, but I felt like doing so today :wink:

Aside from providing the laws, I changed the "left/right unit" terminology to "left/right identity" for monoid, and found some alternative laws for `implies` that result in the same definition, but are actually name-able as they are recognisable as totality and transitivity.

### DIFF
--- a/src/Data/CommutativeRing/Laws.purs
+++ b/src/Data/CommutativeRing/Laws.purs
@@ -1,0 +1,7 @@
+module Data.CommutativeRing.Laws where
+
+import Prelude
+
+-- | `a * b = b * a`
+commutativeMul :: forall a. Eq a => CommutativeRing a => a -> a -> Boolean
+commutativeMul a b = a * b == b * a
--- a/src/Data/EuclideanRing/Laws.purs
+++ b/src/Data/EuclideanRing/Laws.purs
@@ -1,0 +1,39 @@
+module Data.EuclideanRing.Laws where
+
+import Prelude
+
+-- | `one /= zero`
+integralDomainIdentity :: forall proxy a. Eq a => EuclideanRing a => proxy a -> Boolean
+integralDomainIdentity _ = one /= (zero :: a)
+
+-- | if `a` and `b` are both nonzero then so is `a * b`
+integralDomainDivision :: forall a. Eq a => EuclideanRing a => a -> a -> Boolean
+integralDomainDivision a b
+  | a /= zero && b /= zero = a * b /= zero
+  | otherwise = true
+
+-- | For all nonzero `a`, `degree a >= zero`
+nonNegativeDegree :: forall a. Eq a => EuclideanRing a => a -> Boolean
+nonNegativeDegree a
+  | a /= zero = degree a >= zero
+  | otherwise = true
+
+-- | For all `a` and `b`, where `b` is nonzero, let `q = a / b` and
+-- | ``r = a `mod` b``; then `a = q*b + r`, and also either `r = zero` or
+-- | `degree r < degree b`.
+divisionWithRem :: forall a. Eq a => EuclideanRing a => a -> a -> Boolean
+divisionWithRem a b
+  | b /= zero =
+      let
+        q = a / b
+        r = a `mod` b
+      in
+        a == b * q + r && (r == zero || degree r < degree b)
+  | otherwise =
+      true
+
+-- | For all nonzero `a` and `b`, `degree a <= degree (a * b)`
+submultiplicativeDegree :: forall a. Eq a => EuclideanRing a => a -> a -> Boolean
+submultiplicativeDegree a b
+  | a /= zero && b /= zero = degree a < degree (a * b)
+  | otherwise = true
--- a/src/Data/HeytingAlgebra/Laws.purs
+++ b/src/Data/HeytingAlgebra/Laws.purs
@@ -1,0 +1,62 @@
+module Data.HeytingAlgebra.Laws where
+
+import Prelude
+
+import Data.HeytingAlgebra (ff, implies, tt)
+import Laws.Algebraic as Laws
+
+-- | `a || (b || c) = (a || b) || c`
+associativeDisj :: forall a. Eq a => HeytingAlgebra a => a -> a -> a -> Boolean
+associativeDisj = Laws.associative (||)
+
+-- | `a && (b && c) = (a && b) && c`
+associativeConj :: forall a. Eq a => HeytingAlgebra a => a -> a -> a -> Boolean
+associativeConj = Laws.associative (&&)
+
+-- | `a || b = b || a`
+commutativeDisj :: forall a. Eq a => HeytingAlgebra a => a -> a -> Boolean
+commutativeDisj = Laws.commutative (||)
+
+-- | `a && b = b && a`
+commutativeConj :: forall a. Eq a => HeytingAlgebra a => a -> a -> Boolean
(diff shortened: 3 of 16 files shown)


hdgarrood:
Looks great, and I really like the idea of the Laws.Algebraic module, but I worry that using Eq instances and only returning a Boolean may be too limiting: you’re basically going to have to reimplement the same law checks yourself if you want to, say, provide some kind of useful error message upon failure, or use a different method for determining equality (eg, if you’re using Number, you’ll want to be able to provide an epsilon value). Previously: https://github.com/purescript/purescript-prelude/pull/111#issuecomment-275951753

I’d also quite like to stick to our previous policy that we only use ASCII in core libraries

natefaubion:
Is there a reason to have these in `prelude` and not something like a core `laws` library?

hdgarrood:
Yeah, I was previously in favour of having them here as a form of executable documentation, but given that it’s not entirely clear what the best way to encode them is (if a best way even exists), we may well end up wanting to make breaking changes to these and not being able to because it would split the ecosystem (whereas only testing libraries will want to care about these functions, I guess). So I am beginning to think a separate library would indeed be better.

src/Data/EuclideanRing/Laws.purs
```diff
@@ -0,0 +1,39 @@
...
+        r = a `mod` b
+      in
+        a == b * q + r && (r == zero || degree r < degree b)
+  | otherwise =
+      true
+
+-- | For all nonzero `a` and `b`, `degree a <= degree (a * b)`
+submultiplicativeDegree :: forall a. Eq a => EuclideanRing a => a -> a -> Boolean
+submultiplicativeDegree a b
+  | a /= zero && b /= zero = degree a < degree (a * b)
```
hdgarrood, line 38: This does need to be a non strict inequality. Consider the constant polynomials 1 and 2; their degrees are both 1, and their product is again 2 which also has degree 1, but this is a perfectly valid EuclideanRing.

src/Data/CommutativeRing/Laws.purs
```diff
@@ -0,0 +1,7 @@
+module Data.CommutativeRing.Laws where
+
+import Prelude
+
+-- | `a * b = b * a`
+commutativeMul :: forall a. Eq a => CommutativeRing a => a -> a -> Boolean
+commutativeMul a b = a * b == b * a
```
hdgarrood, line 7: Could this use `commutative` from Laws.Algebraic?

src/Data/HeytingAlgebra/Laws.purs
```diff
@@ -0,0 +1,62 @@
...
+
+-- | `a && tt = a`
+identityConj :: forall a. Eq a => HeytingAlgebra a => a -> Boolean
+identityConj = Laws.identity (&&) tt
+
+-- | ``a `implies` a = tt``
+reflexiveImplies :: forall a. Eq a => HeytingAlgebra a => a -> Boolean
+reflexiveImplies a = (a `implies` a) == tt
+
+-- | ``(a && b) `implies` (b && c) = a `implies` c``
```
hdgarrood, line 48: I don’t think we can describe this as transitivity. Usually, saying that a relation ~ is transitive is saying that if A ~ B and B ~ C, then A ~ C. That doesn’t quite match the law here, because && is taking the place of ~ on the left hand side, but `implies` is taking the place of ~ on the right hand side. Also, transitivity doesn’t usually work backwards: knowing that A ~ C doesn’t allow me to deduce that A ~ B and that B ~ C.

src/Data/EuclideanRing/Laws.purs
garyb: Good catch. It even has the right definition in the comment. :wink:

src/Data/HeytingAlgebra/Laws.purs
garyb: You're right, I'm not sure what I was doing here. The way I saw this expressed means this should be:

```
((a `implies` b) `implies` (b `implies` c)) `implies` (a `implies` c) = tt
```

Better?

garyb:
> I’d also quite like to stick to our previous policy that we only use ASCII in core libraries

Oops, that was a mistake.

I'll put these elsewhere, but will leave the PR open for now until the discussion points are resolved.

src/Data/HeytingAlgebra/Laws.purs
hdgarrood: I'm not quite sure about this either: don't we want some of those `implies` to be using the `HeytingAlgebra Boolean` instance, rather than having all of them use the `HeytingAlgebra a` instance?

hdgarrood:
Should we close this? I'm still of the opinion that law functions are likely better off in a separate library, because the specifics of how you decide whether a law is satisfied, as well as how you represent failures, seem to me to be too difficult to do in an unopinionated way.

JordanMartinez:
Would you ever use these laws to test the type class instances within Prelude itself? If not, then I would agree that it should be moved to its own library. If you would, then perhaps it should stay here.

Regardless, it seems like you want a design more like this, which does not include the `Eq` constraint in all situations, and opens up the return result to whatever one wants to do with it.
```purescript
-- Left = failure but stores the original values
-- Right = success and stores the original values
type Result a = Either (Tuple a a) Unit

baseLaw :: forall a. (a -> a -> Boolean) -> a -> a -> Result a a
baseLaw equals left right =
  if equals left right then Right unit else Left (Tuple left right)

baseLawBool :: forall a. (a -> a -> Boolean) -> a -> a -> Boolean
baseLawBool equals left right =
  either (const false) (const true) $ baseLaw equals left right

baseLawEq :: forall a. Eq a => a -> a -> Result a a
baseLawEq = baseLaw (==)

baseLawEqBool :: forall a. Eq a => a -> a -> Boolean
baseLawEqBool = baseLawBool (==)
```

hdgarrood:
I think the question of whether these functions are being used to test the Prelude code shouldn’t be a factor here. If these functions are going to be used for testing the Prelude, then by default they should go in the tests and not be part of the public API. They should only become part of the public API if we expect people to use them downstream. Adding things to the Prelude’s public API should require a very high bar to be cleared, and I don’t think it is in this case.

The design you propose does solve the problems raised, but at that point you’ve pulled so many things out into separate  arguments there’s barely anything left; what you’re left with is so general that I think it’s unlikely to make it any easier for someone else who is trying to write tests for these laws (as opposed to just writing them directly in whatever style is desired in that specific case).

JordanMartinez:
> I think the question of whether these functions are being used to test the Prelude code shouldn’t be a factor here.

Makes sense. I asked my question because if they were packed up into a library, then one would not be able to use them to test Prelude because they would depend on an older version of Prelude than the one someone would be testing.

> The design you propose does solve the problems raised, but at that point you’ve pulled so many things out into separate arguments there’s barely anything left; what you’re left with is so general that I think it’s unlikely to make it any easier for someone else who is trying to write tests for these laws (as opposed to just writing them directly in whatever style is desired in that specific case).

:smile: Ok, so what's the golden mean? I thought that the above `baseLaw` and `baseLawEq` could provide the foundational core, and then another library could build on top of it in an opinionated way.

hdgarrood:
I don’t think there is a golden mean in this case; I don’t think any of the ways of providing law functions in the prelude will be preferable to just letting people express the laws directly in downstream libraries like quickcheck-laws, which can be more opinionated.

JordanMartinez:
Shall we close this PR then?

PR #211, opened 2020-03-29
Fix kind warnings

Thimoteus:


hdgarrood:
These are fixed in the 0.14 branch aren’t they?

JordanMartinez:
@hdgarrood No, not yet. My previous PR came **before** the changes Nate made when core contributors decided to emit the "could not generalize kind" warnings.

src/Data/NaturalTransformation.purs
```diff
@@ -14,6 +14,7 @@ module Data.NaturalTransformation where
 -- | `f` and `g` should be functors, but the `Functor` constraint is not
 -- | enforced here; that the types are of kind `Type -> Type` is enough for our
 -- | purposes.
+type NaturalTransformation :: forall k. (k -> Type) -> (k -> Type) -> Type
```
hdgarrood, line 17: This kind signature disagrees with the doc-comments. I feel like perhaps we should make this `(Type -> Type) -> (Type -> Type) -> Type` unless there is a specific example where the generalized version would be useful.

hdgarrood: (Or alternatively update the doc comments)

Thimoteus: That's a good point, I've updated the doc comments.

hdgarrood:
I don’t remember closing this; I guess must have closed it by accident, sorry!

PR #219, opened 2020-07-18
Beginner-friendly notes on Unit

milesfrain:


src/Data/Unit.purs
```diff
@@ -7,9 +7,15 @@ import Data.Show (class Show)
 -- |
 -- | `Unit` is often used, wrapped in a monadic type constructor, as the
 -- | return type of a computation where only the _effects_ are important.
+-- |
+-- | You can think of `Unit` as being analogous to the `void` keyword in
+-- | other languages, such as C, Java, etc., which indicates that the
+-- | function does not return any _meaningful_ data.
```
milesfrain, line 13: Could mention how `void` can also indicate "no parameters" for a function, but I don't see a lot of uses of `:: Unit -> _` for functions in PS, so I don't think this would be a helpful analogy.

hdgarrood: I'm not sure I'd call this beginner-friendly, it's only really "previous C or Java experience"-friendly. I think this comment could end up being a net negative if enough of the people reading it aren't familiar with C or Java or some other language in which `void` is a keyword. People learning to code today will probably start with JavaScript and/or Python, neither of which have a `void` keyword.

natefaubion: This is a bit "umm actually" 😆 but JavaScript does have a `void` keyword, and it's actually how you get a hold of "undefined" which is really just an identifier.

hdgarrood: Oh lol, yeah, I'd forgotten about JavaScript's.

hdgarrood:
I'm going to close this for now. Thanks anyway for the PR!

PR #225, opened 2020-09-17
Fix CI for preparation for 0.14

hdgarrood:
I'd like to keep CI working while we're preparing the core libraries for the upcoming 0.14.0 release. This seems like a reasonable way of doing it for now. We'll need to go back through the core libraries to tag proper versions again later anyway, so we can revert this change then.

As an alternative I considered making a `v0.14.0-rc2` release on npm, but unfortunately the npm installer isn't happy with that:

```
$ node index.js --purs-ver=0.14.0-rc2
✖ Check if a prebuilt 0.14.0-rc2 binary is provided for linux
  Error: Expected `version` option to be a string of PureScript version, for example '0.12.5', but go
t an invalid version '0.14.0-rc2'.
```

This would probably be relatively easy to fix, but a one-line change in .travis.yml seems easier.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ node_js: stable
 env:
   - PATH=$HOME/purescript:$PATH
 install:
-  - TAG=0.14.0-rc2
+  - TAG=v0.14.0-rc2
   # - TAG=$(basename $(curl --location --silent --output /dev/null -w %{url_effective} https://github.com/purescript/purescript/releases/latest))
   - curl --location --output $HOME/purescript.tar.gz https://github.com/purescript/purescript/releases/download/$TAG/linux64.tar.gz
   - tar -xvf $HOME/purescript.tar.gz -C $HOME/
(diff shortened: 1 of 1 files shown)


thomashoneyman (approved):
We can also make an issue to restore the prior configuration once we’re out of the -rc stage

hdgarrood:
If we are going to make this change in every core library, do you think it would be worth making an issue in every core library?

JordanMartinez:
> do you think it would be worth making an issue in every core library?

As in, an issue to restore the configuration to the original one so we don't forget to do that?

hdgarrood:
Yeah, although I'm not sure we need one. My thinking is that this is going to be an issue for every core library (if we want to be able to use CI while we're updating things for v0.14.0), and we'll need to go back over all of the core libraries anyway once we've released v0.14.0 to cut proper releases and put proper version bounds in their bower.json files, so we could restore the CI configurations then too, and I'm not sure we need an issue to remind us.

thomashoneyman:
If we will already have to back over them for other reasons then let’s skip the issue. It would be useful to have a checklist of updates to make, though, because it’s easy to forget little things like updating the CI file when you’re updating a lot of libraries. Not sure where that should live but I think it would help us catch everything.

PR #229, opened 2020-10-03
Implement `ap` in terms of `Bind`

### LINKED ISSUE #158: Interplay between `Apply` and `Bind`
On the heels of #141, we now have that the `Apply` and `Bind` instances for a data type have to agree when that data type also has a `Monad`. It seems like we're still missing something.

With the laws as they're written, you can have a `Bind` instance for a data type that doesn't agree with its `Apply` instance, so long as said data type does not also define a `Monad` instance. I think that's not right. I think we want to make sure that the `Apply` and `Bind` instances agree.

A real data type that could sneak a lawful `Bind` instance in as the laws are now is `Data.Validation.Semigroup.V`:

```PureScript
instance bindV :: (Semigroup a) => Bind (V a) where
  bind = flip (unV invalid)
```
I'm pretty sure that definition is associative, but it will "short-circuit" at the first invalid case. That doesn't agree with the `Apply` instance (which accumulates invalid cases).

I'm proposing that we change `ap` to something like:

```PureScript
ap :: forall a b f. Bind f => f (a -> b) -> f a -> f b
ap f a = bind f (mapFlipped a)
```

And move the `Applicative Superclass` law to `Bind` (probably change the name of the law as well).

Thoughts?

hdgarrood:
* Moves `ap` to Control.Bind and implements it in terms of just Bind
  rather than Monad
* Moves the Applicative Superclass law from Monad to Bind and renames it
  to the Apply Superclass law

Note that this should be non-breaking since `ap` is now re-exported from
Control.Monad.

Fixes #158

### DIFF
--- a/src/Control/Bind.purs
+++ b/src/Control/Bind.purs
@@ -6,6 +6,7 @@ module Control.Bind
   , composeKleisli, (>=>)
   , composeKleisliFlipped, (<=<)
   , ifM
+  , ap
   , module Data.Functor
   , module Control.Apply
   , module Control.Applicative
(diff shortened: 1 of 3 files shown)


JordanMartinez (approved):

thomashoneyman (approved):

src/Control/Bind.purs
```diff
@@ -134,3 +136,18 @@ infixr 1 composeKleisliFlipped as <=<
 -- | ```
 ifM :: forall a m. Bind m => m Boolean -> m a -> m a -> m a
 ifM cond t f = cond >>= \cond' -> if cond' then t else f
+
+
```
thomashoneyman:
```suggestion
```

PR #230, opened 2020-10-05
Ports `purescript-proxy` into this repo

JordanMartinez:
Since `SProxy` is defined in Prelude, we need to port `purescript-proxy`'s `Proxy` type to here instead now that it can be polykinded.

Note: I assume that `IsSymbol` should use `Proxy` now rather than `SProxy`.


src/Data/Symbol.purs
```diff
@@ -5,20 +5,21 @@ module Data.Symbol
...
 
+import Type.Proxy (Proxy(..))
+
 -- | A value-level proxy for a type-level symbol.
 data SProxy (sym :: Symbol) = SProxy
 
 -- | A class for known symbols
 class IsSymbol (sym :: Symbol) where
-  reflectSymbol :: SProxy sym -> String
+  reflectSymbol :: Proxy sym -> String
```
natefaubion: This can potentially be:

```
reflectSymbol :: forall proxy. proxy sym -> String
```

And old code will continue to work.

src/Data/Symbol.purs
```diff
@@ -8,11 +8,19 @@ module Data.Symbol
...
+-- | **Deprecated as of v0.14.0 PureScript release**: use `Type.Proxy` instead.
 data SProxy (sym :: Symbol) = SProxy
 
 -- | A class for known symbols
 class IsSymbol (sym :: Symbol) where
-  reflectSymbol :: Proxy sym -> String
+  -- Note: Before v0.14.0, we did not have polykinds. Thus, we needed
+  -- kind-specific proxy types to pass a Symbol around.
+  -- Once v0.14.0 was released, we could use the kind-generic `Proxy` type
+  -- insteand. However, to reduce the code breakage, we're using
```
thomashoneyman:
```suggestion
  -- instead. However, to reduce code breakage, we're using
```

src/Data/Symbol.purs
```diff
@@ -8,11 +8,19 @@ module Data.Symbol
...
+  -- `forall proxy. proxy sym` here so that `SProxy` code will still compile.
+  -- When PureScript makes a new breaking release after the v0.14.0 release,
+  -- this type signature will be updated to `Proxy sym -> String`.
```
thomashoneyman, line 23: We should probably make an issue to track this so we don't forget.

src/Data/Symbol.purs
JordanMartinez: Thanks. Fixed in latest commit

JordanMartinez:
I'll have to get to this later:
- Update all kind-specific proxy types to use `forall proxy. proxy kindValue`

src/Data/Symbol.purs
JordanMartinez: Open it up!

src/Data/Symbol.purs
JordanMartinez: Thanks. I forgot about this solution.

src/Data/Symbol.purs
thomashoneyman: @hdgarrood or @natefaubion, do you have suggestions on what that issue ought to touch on?

hdgarrood: Nope, this seems fairly self-explanatory to me. We can always discuss further on the issue too.

src/Data/Symbol.purs
```diff
@@ -5,20 +5,30 @@ module Data.Symbol
...
+  reflectSymbol :: forall proxy. proxy sym -> String
 
 -- local definition for use in `reifySymbol`
 foreign import unsafeCoerce :: forall a b. a -> b
 
-reifySymbol :: forall r. String -> (forall sym. IsSymbol sym => SProxy sym -> r) -> r
-reifySymbol s f = coerce f { reflectSymbol: \_ -> s } SProxy where
+reifySymbol :: forall r. String -> (forall sym. IsSymbol sym => Proxy sym -> r) -> r
```
hdgarrood: I think this should use a type variable `proxy` instead of `Proxy`, for the same reason as `reflectProxy` - that is, so that it works with both Proxy and SProxy until we update it again later.

JordanMartinez: Good catch! I've fixed that in latest commits.

JordanMartinez:
Should the docs for `Proxy` be updated at all? I think right now it only explains how to use it as a proxy for kind `Type`, not for any kinds.

hdgarrood:
I don't think that's necessary; it will be clear that Proxy can be used for any kind once someone has come across the idea of polymorphic kinds, and I think the documentation for polykinds would be a better place to introduce things like the fact that you can use Proxy with any kind.

JordanMartinez:
Ok. Then I think this PR is ready to be merged if there are no other changes I need to make.

hdgarrood (approved):

PR #233, opened 2020-10-11
Fix loop

### LINKED ISSUE #232: Having `ap` depend on `map` can introduce loops
Before #229 was merged, `ap` depended on `bind` and `pure`. Now that #229 has been merged, `ap` depends on `bind` and `map`.

This means that defining `apply = ap` as well as `map = liftA1` will cause loops. When you try to use `map` with such a type, the following happens:

* `map`
* is implemented via `liftA1`,
* which calls `apply`,
* which is implemented via `ap`,
* ... which calls `map`, and we loop.

Effect is one such type; this was causing failures in https://github.com/purescript/purescript-refs/pull/29#issuecomment-706559781.

I think the law changes in #229 were correct, but perhaps changing `ap` to have a `Bind` constraint rather than a `Monad` constraint should be reverted? I think changing the rules for when it's safe to use these default implementation functions is probably too dangerous a breaking change to be justifiable. Maybe default implementation functions should only be allowed to use type class members from classes which are _below_ the class which they are providing a default implementation for in the hierarchy? That is, it's fine for `liftA1` to depend on both `apply` and `pure`, because they are members from `Apply` and `Applicative` respectively, which are both below `Functor`. Also, it's fine for `ap` to depend on both `bind` and `pure`, because they are from `Bind` and `Applicative`, which are both below `Apply`. However, it's not fine for `ap` to depend on `map`, because that's from `Functor`, which is above `Apply`.

JordanMartinez:
Fixes #232

### DIFF
--- a/src/Control/Monad.purs
+++ b/src/Control/Monad.purs
@@ -66,7 +66,7 @@ unlessM mb m =  do
   b <- mb
   unless b m
 
--- | `ap` provides a default implementation of `(<*>)` for any `Bind`, without
+-- | `ap` provides a default implementation of `(<*>)` for any `Monad`, without
 -- | using `(<*>)` as provided by the `Apply`-`Bind` superclass relationship.
 -- |
 -- | `ap` can therefore be used to write `Apply` instances as follows:
--- a/src/Control/Monad.purs
+++ b/src/Control/Monad.purs
@@ -67,7 +67,7 @@ unlessM mb m =  do
   unless b m
 
 -- | `ap` provides a default implementation of `(<*>)` for any `Monad`, without
--- | using `(<*>)` as provided by the `Apply`-`Bind` superclass relationship.
+-- | using `(<*>)` as provided by the `Apply`-`Monad` superclass relationship.
 -- |
 -- | `ap` can therefore be used to write `Apply` instances as follows:
 -- |
--- a/src/Control/Monad.purs
+++ b/src/Control/Monad.purs
@@ -75,6 +75,10 @@ unlessM mb m =  do
 -- | instance applyF :: Apply F where
 -- |   apply = ap
 -- | ```
+-- Note: Only a `Bind` constraint is needed, but this can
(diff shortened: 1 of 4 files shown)


src/Control/Monad.purs
```diff
@@ -64,3 +65,22 @@ unlessM :: forall m. Monad m => m Boolean -> m Unit -> m Unit
 unlessM mb m =  do
   b <- mb
   unless b m
+
+-- | `ap` provides a default implementation of `(<*>)` for any `Bind`, without
```
hdgarrood: “for any Monad”?

JordanMartinez: Forgot to update that.

src/Control/Monad.purs
```diff
@@ -64,3 +65,22 @@ unlessM :: forall m. Monad m => m Boolean -> m Unit -> m Unit
 unlessM mb m =  do
   b <- mb
   unless b m
+
+-- | `ap` provides a default implementation of `(<*>)` for any `Monad`, without
+-- | using `(<*>)` as provided by the `Apply`-`Bind` superclass relationship.
```
hdgarrood: Bind -> Monad here too

JordanMartinez: Why? Isn't this specifically referring to `Apply` and `Bind`? Not `Apply` and `Monad`?

hdgarrood: Because this function has a Monad constraint. It's saying that while you could implement `ap` by just writing `ap = apply` (since the function has a Monad constraint, and Apply is one of Monad's superclasses), we don't, because we want people to be able to write `apply = ap` in their Apply instances.

JordanMartinez: Ah, thanks for clarifying.

JordanMartinez: I've fixed that in latest commits.

hdgarrood (approved):

PR #235, opened 2020-12-19
Port generics-rep to prelude

JordanMartinez:


JordanMartinez:
I'm not sure whether all the data types defined in the Test.GenericsRep are needed anymore since the Enum-related things aren't included here.

test/Test/GenericsRep.purs
```diff
@@ -132,9 +132,7 @@ instance booleanAlgebraB1 :: BooleanAlgebra B1
 
 genericsRepTests :: AlmostEff
 genericsRepTests = do
-  -- Temporarily comment this out to show that CI works
-  -- I'm not sure what this should be
-  -- assert "not sure what this should be" $ show (cons 1 (cons 2 Nil)) == "Cons(1, Cons(2, Nil))"
+  assert "not sure what this should be" $ show (cons 1 (cons 2 Nil)) == "Cons(1, Cons(2, Nil))"
```
JordanMartinez: What should this be? And are there other things that could be affecting this test?

kl0tl: This test used to log `(Cons { head: 1, tail: (Cons { head: 2, tail: Nil }) })` in generics-rep so I think we should assert that this is still the case here.

test/Test/GenericsRep.purs
```diff
@@ -0,0 +1,198 @@
...
+instance ringTuple :: (Ring a, Ring b) => Ring (Tuple a b) where
+  sub (Tuple x1 y1) (Tuple x2 y2) = Tuple (sub x1 x2) (sub y1 y2)
+
+instance heytingAlgebraTuple :: (HeytingAlgebra a, HeytingAlgebra b) => HeytingAlgebra (Tuple a b) where
+  tt = Tuple tt tt
+  ff = Tuple ff ff
+  implies (Tuple x1 y1) (Tuple x2 y2) = Tuple (x1 `implies` x2) (y1 `implies` y2)
+  conj (Tuple x1 y1) (Tuple x2 y2) = Tuple (conj x1 x2) (conj y1 y2)
+  disj (Tuple x1 y1) (Tuple x2 y2) = Tuple (disj x1 x2) (disj y1 y2)
+  not (Tuple x y) = Tuple (not x) (not y)
```
kl0tl: Can’t we use the Option and Pair types we define later instead of redefining Maybe and Tuple here? Perhaps we could then rename Option to Maybe and Pair to Tuple.

test/Test/GenericsRep.purs
```diff
@@ -0,0 +1,198 @@
+module Test.GenericsRep where
```
kl0tl: Should we name this module Test.Data.Generic.Rep, since we’re testing generic implementations based on Data.Generic.Rep?

test/Test/Main.purs
```diff
@@ -14,13 +14,9 @@ main = do
     testIntDivMod
     testIntDegree
     testRecordInstances
+    genericsRepTests
```
kl0tl: Perhaps rather testGenericInstances, to follow the naming scheme of the other test suites?

test/Test/Utils.purs
```diff
@@ -0,0 +1,10 @@
+module Test.Utils where
```
kl0tl, line 1: Test.Assert would be nice here, since this module re-implements the Test.Assert.assert function but this is rather unimportant.

src/Data/Show/Generic.purs
```diff
@@ -0,0 +1,54 @@
...
+  genericShow' (Constructor a) =
+      case genericShowArgs a of
+        [] -> ctor
+        args -> "(" <> intercalate " " ([ctor] <> args) <> ")"
+    where
+      ctor :: String
+      ctor = reflectSymbol (Proxy :: Proxy name)
+
+      intercalate :: forall b. String -> Array b -> String
+      intercalate _ _ = ""
```
kl0tl: I don’t think we can implement this function with only the prelude, it would have to be implemented in JavaScript right?

test/Test/GenericsRep.purs
JordanMartinez: Seems we don't need a Maybe in the first place. I don't recall why I added that but I thought I got an error without it.

I renamed Tuple to Pair and deleted duplicate content. Tests still worked fine after that, so seems good to me.

test/Test/GenericsRep.purs
JordanMartinez: Fixed.

test/Test/Main.purs
JordanMartinez: Fixed.

test/Test/Utils.purs
JordanMartinez: It also has `AlmostEff`, which doesn't have a corresponding entity in `Test.Assert`. I don't think this change needs to be done.

src/Data/Show/Generic.purs
JordanMartinez: Probably.

JordanMartinez: Fixed.

JordanMartinez:
CI builds now.

test/Test/Utils.purs
hdgarrood: I don't think using assert is an option here, but only because assert depends on prelude. Otherwise I'd say this would be worth doing.

hdgarrood:
Sorry to be a pain, but I'm just a bit uncomfortable with discarding previous git history and author information in this way. Can we preserve the previous git history from the other repo for these files?

JordanMartinez:
> I'm just a bit uncomfortable with discarding previous git history and author information in this way. Can we preserve the previous git history from the other repo for these files?

Sure, but I don't know how to do that, nor even where to start.

test/Test/Utils.purs
JordanMartinez: I think kl0tli means renaming the above `Test.Utils` module to `Test.Assert`. However, maybe I'm just misreading his comment?

src/Data/Show/Generic.purs
```diff
@@ -0,0 +1,53 @@
...
+      ctor = reflectSymbol (Proxy :: Proxy name)
+
+instance genericShowArgsArgument :: Show a => GenericShowArgs (Argument a) where
+  genericShowArgs (Argument a) = [show a]
+
+-- | A `Generic` implementation of the `show` member from the `Show` type class.
+genericShow :: forall a rep. Generic a rep => GenericShow rep => a -> String
+genericShow x = genericShow' (from x)
+
+foreign import intercalateImpl :: String -> Array String -> String
```
kl0tl: Don’t we usually drop the Impl suffix from foreign values which are used as-is instead of being used to implement PureScript values?

test/Test/Data/Generics/Rep.purs
```diff
@@ -0,0 +1,187 @@
+module Test.Data.Generics.Rep where
```
kl0tl: The package was named generics-rep but the module is Data.Generic.Rep 😬 Should we rename this module to Test.Data.Generic.Rep instead (and testGenericsRep to testGenericRep)? Perhaps you wanted to emphasize that we’re testing multiple generic implementations though, in which case that’s fine by me.

test/Test/Utils.purs
kl0tl: Yeah, I was suggesting to rename Test.Utils to Test.Assert but this was only to please my aesthetics. Let’s keep Test.Utils if you think this is more appropriate.

src/Data/Show/Generic.purs
JordanMartinez: I don't know. Do we have a documented list of conventions used across core libs?

I'll change the name, but I'd like to understand how to include prior git history from the generics-rep repo first before I make it.

test/Test/Data/Generics/Rep.purs
JordanMartinez: Ah, darn it! Nope, I just spelled it wrong because I was going by the repo name.

kl0tl:
We should be able to preserve the history by creating a branch in generics-rep, preparing the migration with a few commits (renaming the modules, moving test/Main.purs to test/Test/Data/Generic/Rep.purs, dropping the imports of modules not in the prelude, …), filtering the history with [`git filter-branch`](https://git-scm.com/docs/git-filter-branch)

```
git filter-branch -f --tree-filter 'find . -type f ! \( -path "./src/*" -or -path "./test/*" \) -exec rm -rf "{}" \;' --prune-empty
```

then pushing the rewritten branch to this repository.

JordanMartinez:
I'll work on the `generics-rep` branch. I'll post here when I've finished.

JordanMartinez:
Pushed the migration-preparation branch to my own fork of generics-rep: https://github.com/JordanMartinez/purescript-generics-rep/tree/prepareForPreludePort

I'll now run the command and push the resulting branch here.

JordanMartinez:
CI builds and the git history from `generics-rep` is now included in this PR.

kl0tl (approved):
It’s somewhat unfortunate that generics-rep test suite will first appear in the history as test/Main.purs before being moved to test/Data/Generic/Rep.purs. I’m not sure this is worth fixing though, and thankfully it doesn’t conflict with the prelude test suite which was in test/Test/Main.purs from the start :relieved: Thank you!

JordanMartinez:
> It’s somewhat unfortunate that generics-rep test suite will first appear in the history as test/Main.purs before being moved to test/Data/Generic/Rep.purs. I’m not sure this is worth fixing though, and thankfully it doesn’t conflict with the prelude test suite which was in test/Test/Main.purs from the start

I don't think it's worth fixing. I'm going to merge this.

JordanMartinez:
So, looks like I accidentally renamed the class `GenericRep` to `Generic`. That's why the `Maybe` PR isn't building. I'll submit another PR that fixes this. Sorry about that.

JordanMartinez:
Hm... nevermind. [`Generic` is the correct name](https://pursuit.purescript.org/packages/purescript-generics-rep/6.1.1/docs/Data.Generic.Rep#t:Generic).

PR #236, opened 2020-12-20
Show -0.0 sign

### LINKED ISSUE #152: Should `show (-0.0)` produce "-0.0"?
Currently, `show (-0.0) == "0.0"`, which is arguably wrong, since even though `-0` and `0` do compare equal, they produce different results when dividing by either of them.

kl0tl:
Close https://github.com/purescript/purescript-prelude/issues/152.

### DIFF
--- a/src/Data/Show.js
+++ b/src/Data/Show.js
@@ -5,7 +5,7 @@ exports.showIntImpl = function (n) {
 };
 
 exports.showNumberImpl = function (n) {
-  var str = n.toString();
+  var str = n === 0 && 1 / n === -Infinity ? "-0" : n.toString();
   return isNaN(str + ".0") ? str : str + ".0";
 };
 
--- a/test/Test/Main.js
+++ b/test/Test/Main.js
@@ -19,6 +19,7 @@ exports.testNumberShow = function(showNumber) {
         // Within Int range
         [0.0, "0.0"],
         [1.0, "1.0"],
+        [-0.0, "-0.0"],
         [-1.0, "-1.0"],
         [500.0, "500.0"],
 


hdgarrood:
Ah, I’m more conflicted about this after looking at the diff and realising that JS’s .toString doesn’t include the minus sign for negative zero. I can definitely see this causing breakage at runtime without warning. I’ll see if I can find anything about this in IEEE 754.

hdgarrood:
It turns out IEEE 754 doesn't say anything about functions like `.toString()` for formatting floating point numbers as strings for display.

kl0tl:
If we don’t infringe IEEE 754 by showing positive and negative zeroes differently is there any reason not to do it?

hdgarrood:
I think the reasons not to do it are that a) people may be relying on the current behaviour, and b) there might be a good reason why JS's `.toString()` doesn't distinguish positive and negative zero, and I'd like to look into that first. It's probably worth seeking out some examples of the stance other languages have taken on this question and what conclusions they came to. This is definitely not a decision we'll want to rush; much as I'd prefer this wasn't the case, in practice people will be using `Show Number` for a whole lot of different things beyond just debugging and the repl.

hdgarrood:
https://github.com/rust-lang/rust/issues/20596 and https://github.com/rust-lang/rfcs/issues/1074 are relevant

JordanMartinez:
[This comment in runt-lang/rfcs#1074](https://github.com/rust-lang/rfcs/issues/1074#issuecomment-718243936) shows how most popular languages handle this.

One could get around this problem by specifying a newtype that will always include the sign, but that seems like overkill.

kl0tl:
I suspect that Number.prototype.toString hides the sign of negative zero to not leak its existence to end users? You’re right that people are probably using Show to display numbers in user interfaces (I certainly did it myself 😬). Perhaps we should rather make your Debug class the default in the REPL, show the sign of negative zero there and even deprecate Show if we’re feeling bold.

hdgarrood:
I would love to deprecate Show in favour of Debug eventually, yes.

kl0tl:
Let’s close this for now and reconsider when replacing Show with Debug then?

hdgarrood:
Works for me.

PR #238, opened 2020-12-25
Add toRep

### LINKED ISSUE #246: [generics-rep] Add a `toRep` function
```purescript
toRep :: forall from to. Generic from to => Proxy from -> Proxy to
toRep _ = Proxy
```

JordanMartinez:
Fixes #246

### DIFF
--- a/src/Data/Generic/Rep.purs
+++ b/src/Data/Generic/Rep.purs
@@ -2,7 +2,7 @@ module Data.Generic.Rep
   ( class Generic
   , to
   , from
-  , toRep
+  , repOf
   , NoConstructors
   , NoArguments(..)
   , Sum(..)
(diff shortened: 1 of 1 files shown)


src/Data/Generic/Rep.purs
```diff
@@ -34,3 +37,6 @@ newtype Argument a = Argument a
 class Generic a rep | a -> rep where
   to :: rep -> a
   from :: a -> rep
+
+toRep :: forall from to. Generic from to => Proxy from -> Proxy to
```
kl0tl: Could we reuse `a` and `rep` variables, like in the class declaration?

kl0tl: Also I don’t know about you but I frequently confuse `to` and `from`, and I’m sure having `to :: rep -> a` but `toRep :: Proxy a -> Proxy rep` is going to be even more confusing to me 😬 May I suggest `repOf` instead?

JordanMartinez: Sounds good to me!

JordanMartinez: Fixed.

kl0tl (approved):
Thank you!

thomashoneyman (approved):

PR #239, opened 2020-12-25
Add UsingGeneric newtype

JordanMartinez:
Currently fails to build. Not entirely sure why.

Adds this newtype in preparation for #241 / https://github.com/purescript/purescript-generics-rep/issues/36, which depends on https://github.com/purescript/purescript/issues/3302 getting supported, which has a PR open here: https://github.com/purescript/purescript/pull/3824

### DIFF
--- a/src/Data/Generic/UsingGeneric.purs
+++ b/src/Data/Generic/UsingGeneric.purs
@@ -1,0 +1,59 @@
+module Data.Generic.UsingGeneric where
+
+import Data.Generic.Rep (class Generic)
+import Data.Eq (class Eq)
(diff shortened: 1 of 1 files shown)


kl0tl:
Instances for the UsingGeneric newtype needs more constraints on the generic representation of the underlying type, see https://github.com/purescript/purescript-generics-rep/pull/37.

As Harry said in https://github.com/purescript/purescript-generics-rep/pull/37#issuecomment-613710651 though, I think we should close this until we make more progress on deriving via.

JordanMartinez:
Ok. I'll leave the branch on my repo so I can quickly reopen this PR later.

PR #249, opened 2020-12-26
Change NoConstructors to a newtype over Void

### LINKED ISSUE #247: [generics-rep] newtype NoConstructors = NoConstructors Void
This way you can use `absurd`. Although if we implement https://github.com/purescript/purescript/issues/2922 then this is not necessary.

JordanMartinez:
Fixes #247﻿

### DIFF
--- a/src/Data/Generic/Rep.purs
+++ b/src/Data/Generic/Rep.purs
@@ -10,8 +10,10 @@ module Data.Generic.Rep
   , Argument(..)
   ) where
 
+import Data.Void (Void)
+
 -- | A representation for types with no constructors.
-data NoConstructors
+newtype NoConstructors = NoConstructors Void
 
 -- | A representation for constructors with no arguments.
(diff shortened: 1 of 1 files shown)


kl0tl:
I don’t think we should do this: I doubt being able to eliminate terms of type `NoConstructors` with `absurd` is useful enough when we can already use `unsafeCrashWith "unreachable"` or `unsafeCoerce`.

hdgarrood:
I do think that setting things up so that you don’t need to use unsafe functions is generally worth it, but since this doesn’t appear to be a breaking change we can deal with it after 0.14.0.

kl0tl:
I agree, but here I'd prefer to not commit to this encoding yet so we can try to make some progress on https://github.com/purescript/purescript/issues/2922.

JordanMartinez:
Ok. I'll close this then. Thanks for the feedback.

PR #250, opened 2020-12-26
Complete daniel-chambers' PR adding Show instances to generics rep types

JordanMartinez:
Completes the PR he started in https://github.com/purescript/purescript-generics-rep/pull/11

### DIFF
--- a/src/Data/Generic/Rep.purs
+++ b/src/Data/Generic/Rep.purs
@@ -10,24 +10,45 @@ module Data.Generic.Rep
   , Argument(..)
   ) where
 
+import Data.Semigroup ((<>))
+import Data.Show (class Show, show)
+import Data.Symbol (class IsSymbol, reflectSymbol)
+import Type.Proxy (Proxy(..))
+
 -- | A representation for types with no constructors.
 data NoConstructors
 
 -- | A representation for constructors with no arguments.
 data NoArguments = NoArguments
 
+instance showNoArguments :: Show NoArguments where
+  show _ = "NoArguments"
+
 -- | A representation for types with multiple constructors.
 data Sum a b = Inl a | Inr b
 
+instance showSum :: (Show a, Show b) => Show (Sum a b) where
+  show (Inl a) = "(Inl " <> show a <> ")"
+  show (Inr b) = "(Inr " <> show b <> ")"
+
 -- | A representation for constructors with multiple fields.
 data Product a b = Product a b
+
+instance showProduct :: (Show a, Show b) => Show (Product a b) where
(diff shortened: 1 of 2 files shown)


test/Data/Generic/Rep.purs
```diff
@@ -123,6 +123,9 @@ testGenericRep = do
   assert "Checking show" $
     show (cons 1 (cons 2 Nil)) == "(Cons { head: 1, tail: (Cons { head: 2, tail: Nil }) })"
 
+  assert "Checking show for generic types" $
+    show (G.from $ cons 1 Nil) == "(Inr (Constructor \"Cons\" (Argument { head: 1, tail: Nil })))"
+
```
kl0tl: Could we also test the show output for `G.from Nil` to cover `NoArguments` and the `Inl` constructor, as well as a constructor with multiple arguments to cover `Product`?

src/Data/Generic/Rep.purs
```diff
@@ -10,25 +10,46 @@ module Data.Generic.Rep
...
 
+instance showProduct :: (Show a, Show b) => Show (Product a b) where
+  show (Product a b) = "(Product " <> show a <> " " <> show b <> ")"
+
 -- | A representation for constructors which includes the data constructor name
 -- | as a type-level string.
 newtype Constructor (name :: Symbol) a = Constructor a
 
+instance showConstructor :: (IsSymbol name, Show a) => Show (Constructor name a) where
+  show (Constructor a) = "(Constructor " <> show (reflectSymbol (Proxy :: Proxy name)) <> " " <> show a <> ")"
```
hdgarrood: Can I request that we put a `@` symbol before the constructor name? It wouldn’t be valid syntax (we don’t have explicit type applications) but it does at least clarify that the string isn’t an actual constructor argument for Constructor.

JordanMartinez: Meaning this? Yeah, I think that makes sense.
```diff
- "(Constructor " <> show (reflectSymbol ...
+ "(Constructor @" <> show (reflectSymbol ...
```

hdgarrood: Yeah, exactly that.

test/Data/Generic/Rep.purs
JordanMartinez: Yup. Fixed.

src/Data/Generic/Rep.purs
JordanMartinez: Fixed.

hdgarrood (approved):

PR #252, opened 2021-01-01
Generate CHANGELOG.md file using notes from previous GH releases

JordanMartinez:


CHANGELOG.md
```diff
@@ -0,0 +1,163 @@
...
+
+## [v2.1.0](https://github.com/purescript/purescript-prelude/releases/tag/v2.1.0) - 2016-09-30
+
+- Added `whenM` and `unlessM` variants of `when` and `unless` for situations where the conditional is also in `m`
+
+## [v2.0.0](https://github.com/purescript/purescript-prelude/releases/tag/v2.0.0) - 2016-09-26
+
+**Breaking changes**
+- Fixed `=<<` to be right associative, as originally intended
+- Changed behaviour of `Eq` and `Ord` for `NaN` numeric value #91 (@berdario)
```
thomashoneyman, line 91: Unfortunately, unlike the GitHub releases, the raw CHANGELOG readme won't link things like `#91` for pull requests and `@berdario` for GitHub user profiles.

We can choose to fix these (increasing the workload) or leave them as-is (if required, folks can go look up the pull request or user profile themselves) or decide not to track them (removing these references -- I think this would be losing valuable information).

My gut sense is that we probably should accept that changelogs pre-0.14 should be treated as-is, and we simply try to use the standard format (breaking changes / new features / bugfixes / other) with proper links going forward.

thomashoneyman (commented):
I think this is a nice translation of the GitHub releases into a format we can use to have good changelog files for these libraries going forward.

I don't think it's worth the effort to comb through each changelog to switch it to the breaking / new features / bugfixes / other improvements format, or to fix links and references. Instead, I think it's better to just have the concrete CHANGELOG files in the first place and enforce formatting going forward.

thomashoneyman:
Another thing worth noting: in the `contrib` libraries we also have a pull request template which includes a checklist item reminding the pull request author to add their changes to the CHANGELOG. That way we distribute the work of keeping this file up-to-date and ready to use when it comes time to make a proper release.

See: https://github.com/purescript-contrib/purescript-machines/blob/main/.github/PULL_REQUEST_TEMPLATE.md

I think that we should add this template (or a similar one) at the same time that we add CHANGELOG files.

JordanMartinez:
Adding a `PULL_REQUEST_TEMPLATE.md` sounds good to me. But again, I'd rather we agree what that template file should be before submitting PRs, so that we don't have to fix it again in another round of PRs.

thomashoneyman:
Yea. I would also like weigh-in from other core members. But I suggest that if we do add these CHANGELOG.md files that we also add a pull request template which requests that pull request authors also update the changelog. For context, the template file which is used across the `contrib` organization lives here:

https://github.com/purescript-contrib/governance/blob/main/updater/templates/base/.github/PULL_REQUEST_TEMPLATE.md

JordanMartinez:
I've added the pull request template here. Once we get feedback from other core contributors, I'd like to update my script to copy this across all core libraries.

CHANGELOG.md
f-f: FWIW in Spago we use the same format, and I think it's valuable to track this info even if it's not directly linking. I think there's actually value in leaving such links out, because it makes maintaining the changelog less burdensome (I'd say lowering the treshold to keep this up to date is key here), and the raw file (i.e. when you're not looking at a markdown render) stays readable.

thomashoneyman (approved):

PR #227, opened 2020-09-26
Fix warnings and doc-comments for 0.14

hdgarrood:
Like #226, except that it's based on the previous PRs #209 and #211 which mysteriously got closed - apparently by me but I don't remember doing that - and GitHub won't let me reopen them either. I'd like to merge this one instead of #226 since it preserves @Thimoteus's authorship as well as the doc-comment update in NaturalTransformation.

PR #255, opened 2021-01-11
Write changelog for upcoming release

JordanMartinez:
**Description of the change**

Initial draft for what the v0.5.0 (i.e. `v0.14.0` PS release-compatible) release notes should be.

Part of purescript/purescript#3985

---

**Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ Documentation improvements:
 - Added `power` example (#253)
 - Clarify `Array`'s do notation and the purposes of `Monoid` and `Semigroup` newtypes (#217)
 - Clarify `Unit` representation in FFI code (#223)
+- Fix typo: 'ommitted' -> 'omitted' (#220)
 
 Other improvements:
 - Migrated to GitHub Actions (#234)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ Breaking changes:
 - Make library compile on `v0.14.0` (#206)
 - `purescript-proxy` was ported to this repo (#230)
 - `purescript-generics-rep` was ported to this repo (#235)
+- Move the `Applicative` Superclass law from `Monad` to `Bind` and rename it
+to the `Apply` Superclass law (#229)
 
 New features:
 - Added `Bounded` instance for records (#208)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,7 @@ Bugfixes:
 - No longer use reference equality check in `Array`'s `Eq` instance because it breaks referential transparency (#187).
 - Fix Ring laws (#228)
 
-Other improvements:
+Documentation improvements:
 - Added `lift2` example using `Maybe` (#213)
 - Added `const` example (#214)
 - Added `power` example (#253)
@@ -37,6 +37,8 @@ Other improvements:
     - Clarify `Array`'s "do notation"
     - Clarify purpose of `Monoid` and `Semigroup` newtypes
 - Clarify `Unit` representation in FFI code (#223)
+
+Other improvements:
 - Migrated to GitHub Actions (#234)
 - Added a `CHANGELOG.md` file and PR template (#254)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ Bugfixes:
 
 Other improvements:
 
-## [v5.0.0](https://github.com/purescript/purescript-prelude/releases/tag/v4.1.0) - 2021-MONTH-DATE
+## [Unreleased] - YEAR-MONTH-DATE
 
 Breaking changes:
 - Support compiler version `v0.14.0`, and drop support for previous versions (#206, #226)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,8 @@ Other improvements:
 - Migrated to GitHub Actions (#234)
 - Added a `CHANGELOG.md` file and PR template (#254)
 
+## [v4.1.1](https://github.com/purescript/purescript-prelude/releases/tag/v4.1.1) - 2019-04-20
+
 ## [v4.1.0](https://github.com/purescript/purescript-prelude/releases/tag/v4.1.0) - 2018-07-17
 
 - Added `Ord` instance for records
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,9 +34,7 @@ Documentation improvements:
 - Added `lift2` example using `Maybe` (#213)
 - Added `const` example (#214)
 - Added `power` example (#253)
-- Improve documentation in various things (#217)
-    - Clarify `Array`'s "do notation"
-    - Clarify purpose of `Monoid` and `Semigroup` newtypes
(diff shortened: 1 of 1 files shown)


JordanMartinez:
I'm assuming we'll want a meta-issue to track which repos have had their changelog written.

hdgarrood:
I've just spotted that this repo has a v4.1.1 release which isn't mentioned in the changelog, that should probably be fixed. These release notes also don't mention #229, are you still working on them?

JordanMartinez:
No, I didn't include #229 because I believe we reverted that change in #233 / #232.

I'll add the 4.1.1 release. It wasn't included because it was just a tag, not an official GH release.

JordanMartinez:
Should we remove [`unsafeCompare`](https://github.com/purescript/purescript-prelude/commit/d84c89f478f2e9b4a13847b236f0e4cd4a5783b9)? It was deprecated in `v4.1.1`.

hdgarrood:
We reverted the change to `ap` but we did still move the Applicative superclass law from Monad to Bind.

JordanMartinez:
Ah... Gotcha. I've added that to this PR.

CHANGELOG.md
```diff
@@ -12,6 +12,39 @@ Bugfixes:
...
+- Fix Ring laws (#228)
+
+Other improvements:
+- Added `lift2` example using `Maybe` (#213)
+- Added `const` example (#214)
+- Added `power` example (#253)
+- Improve documentation in various things (#217)
+    - Clarify `Array`'s `do notation`
+    - Clarify purpose of `Monoid` and `Semigroup` newtypes
+- Clarify `Unit` representation in FFI code (#223)
```
milesfrain: I don't know if this level of detail (where each PR is mentioned) is necessary for changelogs. It would probably be fine to just write something like "Added more examples" or not even mention that, since it doesn't really affect existing users of the library.
I think the main purpose of the changelog is to help folks quickly review API changes since their last update.

JordanMartinez: I'm good with making this change if others agree. I thought "Added more docs" or something wasn't that helpful either.

thomashoneyman: I think it's reasonable to point out documentation updates, but I agree they ought to be as concise as possible. For example:

```
Other improvements:
- Added examples to documentation comments for `lift2`, `const`, and `power` (#213, #214, #253)
- Improved documentation for `do` notation for `Array`, the purpose of the `Monoid` and `Semigroup` newtypes, and the representation of `Unit` in FFI code (#217, #223)
```

thomashoneyman: (They're already de-prioritized by being bundled under the "Other improvements" section, which is easier to skip if you're only skimming for API changes)

thomashoneyman (approved):
I think others should have a chance to read this changelog and weigh in on its contents, so we should wait for a second review before merging. That said, a 👍 as far as I'm concerned and thank you for getting this started.

CHANGELOG.md
hdgarrood: I disagree, I think almost every PR should be referenced in the changelog. It's not as if there's a massive amount of detail to wade through here, and it's important that we provide credit to contributors for the changes of theirs we've merged.

CHANGELOG.md
```diff
@@ -12,6 +12,39 @@ Bugfixes:
...
+Bugfixes:
+- No longer use reference equality check in `Array`'s `Eq` instance because it breaks referential transparency (#187).
+- Fix Ring laws (#228)
+
+Other improvements:
+- Added `lift2` example using `Maybe` (#213)
+- Added `const` example (#214)
+- Added `power` example (#253)
+- Improve documentation in various things (#217)
+    - Clarify `Array`'s `do notation`
```
hdgarrood: The phrase "do notation" isn't code, so we shouldn't use backticks for it here.

CHANGELOG.md
```diff
@@ -12,6 +12,39 @@ Bugfixes:
 
 Other improvements:
 
+## [v5.0.0](https://github.com/purescript/purescript-prelude/releases/tag/v4.1.0) - 2021-MONTH-DATE
+
+Breaking changes:
+- Make library compile on `v0.14.0` (#206)
```
hdgarrood: I would suggest phrasing this as "Support compiler version `v0.14.0`, and drop support for previous versions", as it's a bit more clear that `v0.14.0` refers to the compiler version that way as opposed to the version of some other tool, and also it means that it's clear that this version of the library can't be used with older compilers.

CHANGELOG.md
hdgarrood: Personally I think having separate bullet points for each of the separate documentation PRs (as you do currently) is ideal. We can group them all under a "Documentation improvements" heading so that people who aren't interested in anything that doesn't affect the actual API can easily skim past them.

hdgarrood:
Let's remove unsafeCompare, it has already been deprecated for over a year.

CHANGELOG.md
thomashoneyman: I agree almost all pull requests should be referenced in the changelog. I'm also fine with bullet points for each documentation PR.

As a note, the more concise version I added doesn't drop any PRs, it just allows some of them to be included in the same line if they're logically related. I also think _that's_ fine, but we should probably have an agreed-upon default.

JordanMartinez:
@hdgarrood I've addressed the feedback you raised.

JordanMartinez:
Any other feedback? Or am I clear to merge this?

CHANGELOG.md
```diff
@@ -12,6 +12,41 @@ Bugfixes:
...
+- No longer use reference equality check in `Array`'s `Eq` instance because it breaks referential transparency (#187).
+- Fix Ring laws (#228)
+
+Documentation improvements:
+- Added `lift2` example using `Maybe` (#213)
+- Added `const` example (#214)
+- Added `power` example (#253)
+- Improve documentation in various things (#217)
+    - Clarify `Array`'s "do notation"
+    - Clarify purpose of `Monoid` and `Semigroup` newtypes
```
hdgarrood: I think having a bullet point labelled "improve documentation in various things" under a parent "Documentation improvements" is a bit redundant. How about:
```suggestion
- Clarify `Array`'s do notation and the purposes of `Monoid` and `Semigroup` newtypes (#217)
```

hdgarrood:
I would like to go over the commits since 4.1.1 just to make sure we haven't missed anything out, but other than that, this looks good to me. Thanks!

hdgarrood:
I think the following PRs should be referenced in this changelog: #220, #227, #238

JordanMartinez:
Added those PRs. How about now?

CHANGELOG.md
```diff
@@ -12,6 +12,41 @@ Bugfixes:
 
 Other improvements:
 
+## [v5.0.0](https://github.com/purescript/purescript-prelude/releases/tag/v4.1.0) - 2021-MONTH-DATE
```
hdgarrood: Oh, last thing, maybe it would be good to leave this stuff under "Unreleased" for now so that we don't forget to update the placeholder link and date? I don't really mind though.

hdgarrood (approved):

CHANGELOG.md
JordanMartinez: Makes sense to me. I'll add that change and then merge.

PR #267, opened 2021-08-02
Change Unit representaiton from `{}` to `undefined`

### LINKED ISSUE #266: Make `unit`'s JS representation `undefined`, not `{}`
## Context

I was working on updating `purescript-cypress` to work with `purescript-spec-mocha`, so that something like this works:
```purescript
main :: Effect Unit
main = runMocha do
  describe "example to-do app" do
    it "should work" do
      flip runReaderT cy do
         void $ visit "/todo"
```
After bundling the file via `spago bundle-app` and running the resulting `js` file via Cypress, Cypress does not run the test at all and instead emits this error:

---

Cypress detected that you invoked one or more cy commands but returned a different value.

The returned value was:

> {}

Because cy commands are asynchronous and are queued to be run later, it doesn't make sense to return anything else.

For convenience, you can also simply omit any return value or return `undefined` and Cypress will not error.

In previous versions of Cypress we automatically detected this and forced the cy commands to be returned. To make tings less magical and clearer, we are now throwing an error. [Learn more](https://docs.cypress.io/guides/references/error-messages#Cypress-detected-that-you-invoked-one-or-more-cy-commands-in-a-custom-command-but-returned-a-different-value)

---

This error only occurs if I use `purescript-spec-mocha` to generate the `describe` and `it` blocks around the actual tests. It's very possible that `purescript-spec-mocha` (or a fork) could handle its FFI for `itAsync` a bit differently.

Regardless of why, should `unit` still be represented as `{}`? Or should it be represented as `undefined`? See also #223.

I can't recall where, but I know in FFI somewhere, we used to do `return {};` for a function that returned unit. We later dropped that as we discovered that the object creation slowed down the FFI.

JordanMartinez:
**Description of the change**

Fixes #266. Did we want to change the wording for the docs at all? Perhaps just say "don't return a value at all" instead?

---

**Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ New features:
 Bugfixes:
 
 Other improvements:
+- Changed `unit`'s FFI representation from `{}` to `undefined` (#267 by @JordanMartinez)
 
 ## [v5.0.1](https://github.com/purescript/purescript-prelude/releases/tag/v5.0.1) - 2021-05-11
 
--- a/src/Data/Unit.js
+++ b/src/Data/Unit.js
@@ -1,3 +1,3 @@
 "use strict";
 
-exports.unit = {};
+exports.unit = undefined;


hdgarrood:
I think the docs probably ought to be changed, but that requires us to make a decision on the question I posed on the issue (in short: is the runtime representation of Unit part of the public API or not). In the meantime I think this can go in.

paf31:
This means you won’t be able to tell if a field in a record with type unit field is present or not, surely?

garyb:
> This means you won’t be able to tell if a field in a record with type unit field is present or not, surely?

It's still possible to tell - `hasOwnProperty` and `"field" in record` both do the right thing, but it's a good point, we'll need to check the implementation is not doing `!!record["field"]` or equivalent things to check for the presence of keys.

- [x] Compiler
- [x] `Record`
- [x] `Object`

JordanMartinez:
@garyb Could you explain the compiler check and where specifically we'd need to look?

garyb:
I've just had another look myself. Pretty sure the compiler-generated code is all good, the only compiler-generated code for records that I can think of is constructing records, and modifying fields in records. When modifying fields of open records it does use `hasOwnProperty`, and for closed records it assigns the unchanged properties with whatever value is in the source.

I also did a search for references to undefined, and as far as I could see we're just inserting or erasing undefineds directly from the AST, never making equality comparisons with them.

JordanMartinez:
When I searched for `unit` throughout the codebase, I only came across [`unit`](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Sugar/TypeClasses.hs#L305-L306), which is only used [`typeClassDictionaryDeclaration`](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Sugar/TypeClasses.hs#L276). That `unit` is a completely separate thing, right?

garyb:
Oh, interesting. Yes and no... that `unit` value is the `SourceType` of `{}`. I think it's something that only exists to make the dictionary constructor code generation work correctly though... so actually, I guess yeah we can say that's entirely separate, but maybe we should name this something else. This is most likely from the early days before we erased "empty" arguments, and at some point a `{}` would actually have been passed in here.

JordanMartinez:
Can I get an approval, so I can merge this? Seems like all issues have been addressed.

hdgarrood (approved):

JordanMartinez:
Thanks!

PR #268, opened 2021-09-11
Add `mapRecord` to simulate Functor Record instance

JordanMartinez:
**Description of the change**

See https://github.com/natefaubion/purescript-tidy-codegen/issues/3#issuecomment-917422266 and its surrounding context.

---

**Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ Notable changes to this project are documented in this file. The format is based
 Breaking changes:
 
 New features:
+- Added `mapRecord` to simulate a `Functor` instance for `Record` (#268 by @JordanMartinez)
+
+  A `Record`'s kind signature (`Row Type -> Type`) does not satisfy the kind signature requirements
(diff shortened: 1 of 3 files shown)


JordanMartinez:
Note: AFAICT, we actually can't define a `Functor` instance for a `Record` for two reasons.
First, because it's kind signature doesn't match the requirements set by `Functor`: `Row Type -> Type` /= `Type -> Type`
Second, because the helper type class `FunctorRecord` needs to know what the `(a -> b)` function's `a` and `b` types are, so it can compute what the output row is. For example
```purescript
instance 
  ( Row.Cons sym (f a) others row
  , Row.Cons sym (f b) newOthers newRow
  ) => ...
```

JordanMartinez:
`record-extra` already defines `mapRecord` and can be used to achieve our ultimate goals. See https://github.com/natefaubion/purescript-tidy-codegen/issues/3#issuecomment-917428698

PR #269, opened 2021-09-19
Add nub constraint on Record Show instance

### LINKED ISSUE #216: Record instances using unsafe functions break in the presence of duplicate labels.
For reference: https://github.com/purescript/purescript-record/issues/59

If we want to continue using the unsafe variations instead of moving more of `purescript-record` into Prelude, then we can require a `Nub` constraint, and `RowList` that.

JordanMartinez:
**Description of the change**

Fixes #216. I'm not sure if this would be considered breaking or not.

---

**Checklist:**


JordanMartinez:
Looks like this can't be fixed. CI fails with:
```

[1/1 NoInstanceFound] src/Data/Show.purs:50:22

  50    show record = case showRecordFields (Proxy :: Proxy ls) record of
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  
  No type class instance was found for
  
    Data.Show.ShowRecordFields ls3
                               rs4
  
  while applying a function showRecordFields
    of type ShowRecordFields t0 t1 => t2 t0 -> Record t1 -> Array String
    to argument Proxy
  while inferring the type of showRecordFields Proxy
  in value declaration showRecord
  
  where ls3 is a rigid type variable
          bound at (line 0, column 0 - line 0, column 0)
        rs4 is a rigid type variable
          bound at (line 0, column 0 - line 0, column 0)
        t1 is an unknown type
        t0 is an unknown type
        t2 is an unknown type

           Src   Lib   All
Warnings   0     0     0  
Errors     1     0     1  
```

hdgarrood:
I think saying "this can't be fixed" is a bit premature. It looks to me like you're (correctly) asking for a Nub constraint as well as the an ShowRecordFields instance for `rs'` in the Show instance context, but in the actual implementation you're still asking for a ShowRecordFields instance for `rs`, which you now don't have.

JordanMartinez:
Ah... That explains the type error.

JordanMartinez:
I added a second `Nub` constraint, but thinking about this more, couldn't we also wrote `Union () rs' rs`?

JordanMartinez:
:ping_pong: @thomashoneyman

thomashoneyman:
This one's a bit out of my wheelhouse — perhaps @hdgarrood could take another look?

src/Data/Show.purs
```diff
@@ -41,7 +42,12 @@ instance showArray :: Show a => Show (Array a) where
 instance showProxy :: Show (Proxy a) where
   show _ = "Proxy"
 
-instance showRecord :: (RL.RowToList rs ls, ShowRecordFields ls rs) => Show (Record rs) where
+instance showRecord ::
+  ( Nub rs rs'
+  , Nub rs' rs
```
natefaubion: Can this be asserted with `Nub rs rs`?

hdgarrood (approved):
Seems reasonable to me. I think this is breaking - if you are using `show` on a record whose type isn't concrete, then you might not have a `Nub rs rs` instance in scope and presumably your code would break. For example, if for some reason you wrote

```purescript
myShowRecord :: forall rs ls. (RL.RowToList rs ls, ShowRecordFields ls rs) => Record rs -> String
myShowRecord = show
```
then I think that would break after this change.

JordanMartinez:
So...
1. Should I update the Changelog file to remove the bugfix entry?
2. Should I mention Harry's comment as description in the breaking change entry?

hdgarrood:
I think removing the bugfix entry from changelog makes sense, but I'd say my previous comment is more detail than is appropriate for the changelog.

PR #270, opened 2021-09-19
Add clearer docs for Prelude module

### LINKED ISSUE #251: Add high-level documentation
This issue was raised in a [slack conversation](https://functionalprogramming.slack.com/archives/C717K38CE/p1609215259488600): 

> where is it documented  what Prelude is
I just started and I see Prelude import as the first thing in most file
but I cant find a documentation that explains that

> Looks like it's only mentioned here with the hope that readers will jump to the appropriate conclusions about what it is, although I agree that's not ideal. https://github.com/purescript/documentation/blob/master/guides/Getting-Started.md
It would be better to add some details here: https://pursuit.purescript.org/packages/purescript-prelude/4.1.1
or here:
https://pursuit.purescript.org/packages/purescript-prelude/4.1.1/docs/Prelude

JordanMartinez:
**Description of the change**

Fixes #251

---

**Checklist:**

### DIFF
--- a/src/Prelude.purs
+++ b/src/Prelude.purs
@@ -1,3 +1,15 @@
+-- | `Prelude` is a module that re-exports many other foundational modules from the `purescript-prelude` library
+-- | (e.g. the Monad type class hierarchy, the Monoid type classes, Eq, Ord, etc.).
+-- |
+-- | Typically, this module will be imported in most other libraries and projects as an open import.
+-- |
+-- | ```
+-- | module MyModule where
+-- |
+-- | import Prelude -- open import
+-- |
+-- | import Data.Maybe (Maybe(..)) -- closed import
+-- | ```
 module Prelude
   ( module Control.Applicative
   , module Control.Apply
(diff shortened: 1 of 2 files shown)


src/Prelude.purs
```diff
@@ -1,3 +1,15 @@
...
+-- | (e.g. the Monad type class hierarchy, the Monoid type classes, Eq, Ord, etc.).
+-- |
+-- | Typically, this module will be imported in most other libraries and projects as an open import.
+-- |
+-- | ```
+-- | module MyModule where
+-- |
+-- | import Prelude -- open import
+-- |
+-- | import Data.Maybe (Maybe(..)) -- closed import
```
thomashoneyman, line 11: I’m not sure we need to describe this as a closed import, or comment next to the Prelude import as an open one, but I don’t feel too strongly about it

thomashoneyman (approved):

src/Prelude.purs
JordanMartinez: If the 'open import' comment confuses people, we can see how to edit this later. For now, I think it does more good than harm.

PR #274, opened 2021-09-26
Simplify implementation of internal FFI function intercalate in Data.Show.Generic

cdepillabout:
**Description of the change**

The `Data.Show` and `Data.Show.Generic` modules both use an internal FFI helper function `intercalate`.  It was called `join` in `Data.Show`.

This PR does the following:

- Simplifies the implementation in `Data.Show.Generic.intercalate`.  This now matches the same (simple) implementation from `Data.Show`.
- Renames the function in `Data.Show` to be `intercalate`.  My reasoning is that the function should be named the same thing in both `Data.Show` and `Data.Show.Generic`.  I thought `intercalate` would be easier to understand for a PureScriter, since `intercalate` is already used in modules like `Data.Foldable`.  (Alternatively, `intercalate` could be renamed `join` in both modules, since `join` is the name of the underlying FFI function.  I decided against this since `join` has a different meaning in PureScript. I thought `intercalate` was easier to understand.)

---

**Checklist:**


cdepillabout:
I didn't add a changelog entry for this, since this change shouldn't be visible to any users.

thomashoneyman:
@cdepillabout This can be added to the CHANGELOG under the "Other improvements" section if you would like, but it isn't necessary if you don't want to.

thomashoneyman (approved):

JordanMartinez (approved):
Perhaps a changelog should be added for other backend maintainers because the function name is changing from `join` to `intercalate`?

cdepillabout:
Thanks for the review.

I've added a CHANGELOG entry for this change.  Hopefully this should be good to merge now.

JordanMartinez:
I'd like to merge this, but I'll clarify how this change affects non-JS backends before doing so. While this isn't a breaking change for the JS backend, I wonder if it is for non-JS backends.

JordanMartinez:
Per Nate, this counts as a breaking change for non-JS backends. So, we shouldn't merge it until we're making breaking changes in general.

I've added that label here.

JordanMartinez:
:ping_pong: @thomashoneyman

thomashoneyman (approved):

PR #280, opened 2022-02-13
Fix signum implementation

### LINKED ISSUE #263: `signum` should return `zero` when given `zero` (currently returns `one`)
So I asked about this on the Slack channel, and @hdgarrood said that this was probably an oversight and that he thinks this should be fixed, based on the assumption that the function is not widely used, as this question was not asked before as far as he is aware.

JordanMartinez:
**Description of the change**

Fixes #263. This is a breaking change that shouldn't be merged until we start the v0.15.x release cycle.

---

**Checklist:**


src/Data/Ord.purs
```diff
@@ -214,10 +214,12 @@ between low hi x
 abs :: forall a. Ord a => Ring a => a -> a
 abs x = if x >= zero then x else negate x
 
--- | The sign function; always evaluates to either `one` or `negate one`. For
--- | any `x`, we should have `signum x * abs x == x`.
+-- | The sign function; always evaluates to `negate one`, `zero` or `one`.
```
hdgarrood: This comment doesn't really tell you what the function does. Also I think the second sentence can stay - that should still be true, surely?

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ Notable changes to this project are documented in this file. The format is based
 ## [Unreleased]
 
 Breaking changes:
+- Fix `signum` to return `one` when `x > zero` (#280 by @JordanMartinez)
```
hdgarrood:
```suggestion
- Fix `signum zero` to return `zero` (#280 by @JordanMartinez)
```

src/Data/Ord.purs
hdgarrood:
```suggestion
-- | The sign function; returns `one` if the argument is positive, `negate one` if the argument is negative, or `zero` if the argument is `zero`. For any `x`, we should have `signum x * abs x == x`.
```

src/Data/Ord.purs
```diff
@@ -214,10 +214,12 @@ between low hi x
...
 signum :: forall a. Ord a => Ring a => a -> a
-signum x = if x >= zero then one else negate one
+signum x =
+  if x < zero then negate one
+  else if x > zero then one
+  else zero
```
hdgarrood:
```suggestion
  else x
```
Returning `x` rather than `zero` ought to match the behaviour of `Math.sign` for `Number`.

JordanMartinez: To clarify, will `signum -0.0` return -1 and `signum 0.0` return `zero`?

JordanMartinez: I've merged this change, but just wanted to clarify.

src/Data/Ord.purs
```diff
@@ -214,10 +214,12 @@ between low hi x
 abs :: forall a. Ord a => Ring a => a -> a
 abs x = if x >= zero then x else negate x
 
--- | The sign function; always evaluates to either `one` or `negate one`. For
--- | any `x`, we should have `signum x * abs x == x`.
+-- | The sign function; returns `one` if the argument is positive, `negate one` if the argument is negative, or `zero` if the argument is `zero`. For any `x`, we should have `signum x * abs x == x`.
```
triallax: Since `signum` now returns `x` instead of `zero`, maybe the docs should mention this?

```suggestion
-- | The sign function; returns `one` if the argument is positive, `negate one` if the argument is negative, or `x` if the argument is `zero`. For any `x`, we should have `signum x * abs x == x`.
```

Also, maybe the comment should be broken into multiple lines? This is both more consistent with the rest of the file and easier to read.

src/Data/Ord.purs
triallax: I think @hdgarrood was referring to the value of `Math.sign(-0.0)` in JS being `-0`.

hdgarrood: Yeah, we want `signum (-0.0) = (-0.0)` and `signum 0.0 = 0.0`. Possibly worth adding a test so that this is a bit clearer?

src/Data/Ord.purs
hdgarrood: While this is more accurate, I think it’s also likely to seem a weird way of wording it, especially if people aren’t thinking of Number or aren’t aware of how floats have signed zeroes. Also it might not be clear to people who are reading HTML docs what `x` is here.  Maybe it’d be better to leave this sentence as is and add another one at the end saying that for floating point numbers with signed zeroes, when called with a zero, this function returns the argument in order to preserve the sign?

JordanMartinez: How's the new rendering?

src/Data/Ord.purs
JordanMartinez: Added.

test/Test/Main.purs
```diff
@@ -151,3 +152,8 @@ testRecordInstances = do
   assert "Record top" $
     (top :: { a :: Boolean }).a
     == top
+
+testSignum :: AlmostEff
+testSignum = do
+  assert "signum positive zero" $ signum 0.0 == 0.0
+  assert "signum negative zero" $ signum (-0.0) == (-0.0)
```
hdgarrood: This doesn't work unfortunately - positive zero compares equal to negative zero. Maybe try `show`-ing the result?

src/Data/Ord.purs
```diff
@@ -214,10 +214,16 @@ between low hi x
 abs :: forall a. Ord a => Ring a => a -> a
 abs x = if x >= zero then x else negate x
 
--- | The sign function; always evaluates to either `one` or `negate one`. For
--- | any `x`, we should have `signum x * abs x == x`.
+-- | The sign function; returns `one` if the argument is positive,
+-- | `negate one` if the argument is negative, or `zero` if the argument is `zero`.
+-- | For floating point numbers with signed zeroes, when called with a zero,
+-- | this function returns the argument in order to preserve the sign
```
hdgarrood:
```suggestion
-- | this function returns the argument in order to preserve the sign.
```

src/Data/Ord.purs
hdgarrood: LGTM :+1:

test/Test/Main.purs
JordanMartinez: Changing the test to
```purescript
assert ("signum negative zero: " <> show (signum (-0.0))) $ show (signum (-0.0)) == "-0.0"
```
produces this error:
```
Error: signum negative zero: 0.0
```

JordanMartinez: And `assert (show (-0.0)) $ false` produces `Error: 0.0`

JordanMartinez: Looks like [`Object.is`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is#using_object.is) is a way to verify this.

test/Test/Main.purs
```diff
@@ -155,5 +155,10 @@ testRecordInstances = do
 
 testSignum :: AlmostEff
 testSignum = do
-  assert "signum positive zero" $ signum 0.0 == 0.0
-  assert "signum negative zero" $ signum (-0.0) == (-0.0)
+  assert "signum positive zero" $ objectIs (signum 0.0) 0.0
+  assert "signum negative zero" $ objectIs (signum (-0.0)) (-0.0)
+
+-- Seems to be only way to check whether zero is `-0.0` or `0.0`
```
hdgarrood: I forgot that our `show` doesn't distinguish negative and positive zero. But, you can still distinguish them without the FFI:
```
show (1.0/0.0) == "Infinity"
show (1.0/(-0.0)) == "-Infinity"
```

JordanMartinez:
:ping_pong: @thomashoneyman

thomashoneyman (approved):

PR #281, opened 2022-02-15
Replace polymorphic proxy workaround with monomorphic Proxy type

### LINKED ISSUE #231: Update type signature for `reflectSymbol` for PureScript 0.15
As discussed in #230, we need to change this type signature:

https://github.com/purescript/purescript-prelude/blob/11004596cbaaa7f09274fa4c8554e1c0631755a6/src/Data/Symbol.purs#L23

...when PureScript 0.15 comes out. The signature is using a type variable `proxy` instead of `Proxy` just to avoid too much breakage in 0.14, but we'll want to update it in the next compiler version.

JordanMartinez:
**Description of the change**

Fixes #231 by changing `forall proxy. proxy SomeKind -> ..` to just `Proxy SomeKind`.

Also removes `Proxy2` and `Proxy3` types since `Proxy` is enough now.

---

**Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ Notable changes to this project are documented in this file. The format is based
 
 Breaking changes:
 - Migrated FFI to ES Modules (#287 by @kl0tl and @JordanMartinez)
+- Replace `forall proxy. proxy k` workaound with `Proxy k` (#281 by @JordanMartinez)
+- Drop all kind-specific Proxy types (e.g. `Proxy2`, `Proxy3`) (#281 by @JordanMartinez)
 
 New features:
 
--- a/src/Control/Bind.purs
+++ b/src/Control/Bind.purs
@@ -1,10 +1,16 @@
 module Control.Bind
-  ( class Bind, bind, (>>=)
-  , bindFlipped, (=<<)
-  , class Discard, discard
+  ( class Bind
+  , bind
+  , (>>=)
+  , bindFlipped
+  , (=<<)
+  , class Discard
+  , discard
   , join
-  , composeKleisli, (>=>)
-  , composeKleisliFlipped, (<=<)
+  , composeKleisli
+  , (>=>)
+  , composeKleisliFlipped
+  , (<=<)
   , ifM
   , module Data.Functor
   , module Control.Apply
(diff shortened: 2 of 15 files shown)


garyb:
Is there a downside to leaving `proxy` polymorphic?

JordanMartinez:
I'm not sure. I suppose one could do something like 
```purescript
data MyProxyType (sym :: Symbol) = MyProxyType OtherInfo

reflectSymbol (MyProxyType "foo" :: MyProxyType "bar")
```
But if this is hard-coded to `Proxy`, I don't think it would cause problems either.

thomashoneyman:
> Is there a downside to leaving `proxy` polymorphic?

What about the opposite — is there an advantage to leaving it polymorphic? As far as I remember it was a stopgap to reduce breakage but the plan was to remove it.

The main thing that comes to mind is that leaving it polymorphic is a little less clear than an explicit Proxy argument (to me), including in type errors. As a consequence we have things like Proxy2 or Proxy3, which really aren’t necessary anymore (right?).

garyb:
The hypothetical advantage is you could use types other than proxy that carry the symbol... like, say you have a generic `Id` type that uses a symbol for discriminating between different entities:

``` purescript
newtype Id (tag ∷ Symbol) = Id NonEmptyString

type ItemId = Id "Item"
type OrderId = Id "Order"
```

You could just pass your `ItemId` or `OrderId` to places that currently expect `proxy sym`:

``` purescript
printIdType ∷ ∀ tag. IsSymbol tag ⇒ Id tag → String
printIdType = reflectSymbol
```

But it's not really that big a deal either, it should always be possible to instantiate a `Proxy` with the appropriate type anyway:

``` purescript
printIdType ∷ ∀ tag. IsSymbol tag ⇒ Id tag → String
printIdType _ = reflectSymbol (Proxy ∷ _ tag)
```

At the worst it would require adding some type signatures. Also "constructing" a `Proxy` is just a value dereference at runtime, the same as passing the `Id _` value here, so if we think it's better for understandability purposes to make it monomorphic then I think that's fine.

I do actually have some types like `Id` here, but when I checked I found I'm always using `Proxy` anyway due to the way things worked out. :smile:

thomashoneyman (approved):

PR #283, opened 2022-02-17
Migrate to `purs-changelog` for handling changelog file

JordanMartinez:
**Description of the change**

Uses [`purs-changelog`](https://github.com/purescript-contrib/purescript-up-changelog) to handle the changelog for this repo. (This tool is a port of the `update-changelog.hs` script used in the PureScript repo). The benefit of this change is that we will no longer get merge conflicts in the `CHANGELOG.md` file when merging multiple PRs in a short time period. To see an example of what using `purs-changelog` for updating a changelog looks like, see [this commit](https://github.com/purescript-contrib/purescript-up-changelog/pull/13/commits/3a3e57e46c87cc2406f271845c8f1f9763eccd65)

Moreover, the first checkbox in the PR template list below will no longer require authors to indicate the PR number (so, no more "open a PR, see what the PR number is, and push a commit for that number).

Lastly, when we need to update the changelog for a new release, one only needs to run `purs-changelog update` so long as the URL for the `origin` remote name (i.e. `git remote -v`) refers to this repo and not a fork.

By making all `core`, `contrib`, `web`, and `node` libraries use this tool, in the future we can use GitHub's API to produce a single file that lists all the changes made in an ecosystem update (e.g. v0.15.x release).

---

**Checklist:**

### DIFF
--- a/CHANGELOG.d/README.md
+++ b/CHANGELOG.d/README.md
@@ -1,0 +1,77 @@
+# About
+
+This directory contains changelog entries for work that has not yet been
+released. When a release goes out, these files will be concatenated and
+prepended to `CHANGELOG.md` in a new section for that release.
+
+## For Maintainers
+
+See https://github.com/JordanMartinez/purescript-up-changelog for details of this process.
+
+## For Contributors
+
+When making a new PR, do the following steps. Each is described in the sections that follow:
+1. Add a new file to this directory where the file name follows the naming convention described below
+1. Fill that file with the proper content
+
+To ensure you're doing it correctly, see the [Checklist](#checklist)
+
+### File Naming Convention
+
+The file should be named `{PREFIX}_{SLUG}.md`.
+
+`{PREFIX}` is one of the following:
+* `breaking`: for breaking changes
+* `feature`: for new features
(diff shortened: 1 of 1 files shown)


CHANGELOG.d/README.md
```diff
@@ -0,0 +1,77 @@
+# About
```
thomashoneyman, line 1: I think this file needs to move into the governance or some other centralized repository so that it isn't duplicated in every repo using this model. Instead, this README file can just have a brief description (the paragraph below is fine), and then a link over to the centralized document.

JordanMartinez:
My original goal for the `purs-changelog` project was to be used in single projects, not across whole orgs. I agree with what you're saying [here](https://github.com/purescript/purescript-prelude/pull/283#discussion_r814261804) in that we should have a centralized place for documentation. However, I'm now wondering how I should incorporate that into the design of the tool since that wasn't something I had considered. Moreover, one point you mentioned against doing this was doing this in contrib libraries first as a sort of pilot program before we move it to core in case there are any issues with the code. 

So, for now, I'm going to close this PR.

PR #284, opened 2022-03-02
Update to v0.15.0

JordanMartinez:
**Description of the change**

Backlinking to purescript/purescript#4244

Migrates FFI to ES modules

---

**Checklist:**

- [ ] Added the change to the changelog's "Unreleased" section with a reference to this PR (e.g. "- Made a change (#0000)")
- [x] Linked any existing issues or proposals that this pull request should close
- [ ] Updated or added relevant documentation
- [ ] Added a test for the contribution (if applicable)

PR #264, opened 2021-06-30
Convert FFI to ES modules

i-am-the-slime:
Changes all foreign modules to use ES modules instead of CJS.

**Checklist:**

- [ ] Added the change to the changelog's "Unreleased" section with a reference to this PR (e.g. "- Made a change (#0000)")
- [ ] Linked any existing issues or proposals that this pull request should close
- [ ] Updated or added relevant documentation
- [ ] Added a test for the contribution (if applicable)

PR #287, opened 2022-03-04
Migrate FFI to ES modules

JordanMartinez:
**Description of the change**

Backlinking to purescript/purescript#4244

Migrates FFI to ES modules. This continues what was started in #284 but using a different branch named `es-modules-libraries` to ensure further development on the compiler isn't affected by commits pushed to this PR. 

This PR also succeeds #264 since that PR was not opened by the 'working-group-purescript-es' org, which allows multiple people to push to the PR if the original submitter is gone for while.

This PR will fail to build until `pulp` has been updated to support ES modules when one calls `pulp run` and `pulp test`. See https://github.com/purescript-contrib/pulp/issues/400 for more info.

---

**Checklist:**

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,10 +1,11 @@
 {
   "parserOptions": {
-    "ecmaVersion": 5
+    "ecmaVersion": 6,
+    "sourceType": "module"
   },
   "extends": "eslint:recommended",
   "env": {
-    "commonjs": true
+    "es6": true
   },
   "rules": {
     "strict": [2, "global"],
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -4,9 +4,6 @@
     "sourceType": "module"
   },
   "extends": "eslint:recommended",
-  "env": {
-    "es6": true
-  },
   "rules": {
     "strict": [2, "global"],
     "block-scoped-var": 2,
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,16 +1,23 @@
 {
   "parserOptions": {
-    "ecmaVersion": 5
+    "ecmaVersion": 6,
+    "sourceType": "module"
   },
   "extends": "eslint:recommended",
   "env": {
-    "commonjs": true
+    "es6": true
   },
   "rules": {
-    "strict": [2, "global"],
+    "strict": [
+      2,
+      "global"
+    ],
     "block-scoped-var": 2,
     "consistent-return": 2,
-    "eqeqeq": [2, "smart"],
+    "eqeqeq": [
+      2,
+      "smart"
+    ],
     "guard-for-in": 2,
     "no-caller": 2,
     "no-extend-native": 2,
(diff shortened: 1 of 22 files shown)


JordanMartinez:
PRs to `pulp` to support ES modules are ready for review:
- Migrate `pulp` to GH Actions: https://github.com/purescript-contrib/pulp/pull/399
- Update `pulp` to work on v0.15.0 PS version: https://github.com/purescript-contrib/pulp/pull/401

JordanMartinez:
CI builds now :tada:

JordanMartinez:
The `alpha-02` release is now on GH. Restarting CI to verify this PR still builds.

JordanMartinez:
CI builds on the `alpha-02` release.

thomashoneyman:
We need to be careful that `setup-purescript` has auto-updated to the latest unstable release before re-running CI. For example, it's still on `alpha-01`:

https://github.com/purescript-contrib/setup-purescript/blob/81d7c48c3f39de23dcd3626ab216699010200db2/dist/versions-v2.json#L2-L4

and so that re-run of CI pulled in the old purs:

```
Cached path /opt/hostedtoolcache/purs/0.15.0-alpha-01/x64, adding to PATH
```

I've restarted the sync versions job there so that it updates the file and then we can re-run CI here.

JordanMartinez:
Ah, good catch.

.eslintrc.json
```diff
@@ -1,10 +1,11 @@
 {
   "parserOptions": {
-    "ecmaVersion": 5
+    "ecmaVersion": 6,
+    "sourceType": "module"
   },
   "extends": "eslint:recommended",
   "env": {
-    "commonjs": true
+    "es6": true
```
thomashoneyman: I don't recall if we ended up agreeing on ES6 in the eslint configurations vs. just updating the `parserOptions.sourceType` field. Do you happen to have a link to that discussion?

If we are going forward with ES6, then I'm curious if we need to set both `env: es6` and `parserOptions.ecmaVersion: 6`. What happens if we just remove `env` but still set the parser option to ecmaVersion 6?

thomashoneyman: I don't recall if we ended up agreeing on ES6 in the eslint configurations vs. just updating the `parserOptions.sourceType` field. Do you happen to have a link to that discussion?

If we are going forward with ES6, then I'm curious if we need to set both `env: es6` and `parserOptions.ecmaVersion: 6`. What happens if we just remove `env` but still set the parser option to ecmaVersion 6?

thomashoneyman (approved):

thomashoneyman:
This looks good to me other than the discussion around the eslint configuration.

package.json
```diff
@@ -7,8 +7,8 @@
   },
   "devDependencies": {
     "eslint": "^7.15.0",
-    "purescript-psa": "^0.8.0",
-    "pulp": "^15.0.0",
+    "purescript-psa": "^0.8.2",
+    "pulp": "16.0.0-0",
```
thomashoneyman, line 11: What's the reason for the prerelease version of Pulp? Should this be `16.0.0`?

JordanMartinez: I made a pre-release of `16.0.0` in case we wanted to add support for esbuild later or come across other bugs.

.eslintrc.json
JordanMartinez: Huh... I remember this being important, but I can't recall why. However, running `eslint src test` doesn't report anything with `es6: true` removed.

thomashoneyman:
@JordanMartinez Is it acceptable to delete the es-modules-libraries branch, or does it need to be kept around for the sake of upstream packages?

JordanMartinez:
The `working-group-purescript-es/package-set#main` should be using the `es-modules` branch still, but `pulp@16.0.0-0` would still try to install the branch via `pulp ini`. So, perhaps we should hold off until a `pulp@16.0.0-1` can be released that just installs `master` from the core repo, not the `es-modules-libraries` branch from the working group repo?

PR #293, opened 2022-04-21
Fix String show instance

### LINKED ISSUE #258: Show String outputs illegal escape sequence \& in string literals
## Description

The REPL outputs string literals containing the escape sequence `\&` but this escape sequence is not accepted as input.

## To Reproduce

In the REPL:

```
> "\x0000001"
"\0\&1"

> "\0\&1"
Illegal character escape code at line 1, column 3
```

## Expected behavior

The REPL outputs valid string literals.

## Additional context

Haskell supports `\&` to delimit a numeric escape sequence from the following characters, see [The zero-width escape sequence](http://book.realworldhaskell.org/read/characters-strings-and-escaping-rules.html#id690781). This is exactly the way in which the PureScript REPL uses `\&` to produce canonicalized string literals. However the compiler does not actually accept this escape sequence. If it did this would provide a solution to the problem in purescript/purescript#3750.

## PureScript version

0.13.8

JordanMartinez:
**Description of the change**

Fixes #258 based on [`prettyPrintStringJs`](https://github.com/purescript/purescript/blob/master/lib/purescript-cst/src/Language/PureScript/PSString.hs#L195-L213). I'm not sure whether this implementation is correct or the best.

Also, the test does not currently pass.

```console
# repl
> "\x0000001"
"\x001"
```

CC @MonoidMusician

---

**Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Breaking changes:
 - Replaced polymorphic proxies with monomorphic `Proxy` (#281, #288 by @JordanMartinez)
 - Fix `signum zero` to return `zero` (#280 by @JordanMartinez)
 - Fix `Show` instance on records with duplicate labels by adding `Nub` constraint (#269 by @JordanMartinez)
+- Fix `Show` instance for String (#270 by @JordanMartinez)
 
 New features:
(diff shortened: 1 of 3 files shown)


JordanMartinez:
Looks like this won't get in for `0.15.0`.

PR #295, opened 2022-05-09
Adds yoneda-inspired thunks for Category and Monoid

mikesol:
**Description of the change**

I've written these functions 3-4 times in various repos here and there, and I think they're of enough utility to be in the prelude.

Both of them are yoneda-inspired and basically fish out a value from a closure. For example, `cayo` can be used with profunctor lenses, ie `t = cayo (_2 <<< _2 <<< _2) (1 /\ 2 /\ 3 /\ 4)`.

If folks are ok with this, lemme know and I'll make the relevant changelog entries.

**Checklist:**

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,14 +1,14 @@
 {
-	"private": true,
-	"scripts": {
-		"clean": "rimraf output && rimraf .pulp-cache",
-		"build": "eslint src && pulp build -- --censor-lib --strict",
-		"test": "pulp test"
-	},
-	"devDependencies": {
-		"eslint": "^7.15.0",
-		"purescript-psa": "^0.8.2",
-		"pulp": "^16.0.0-0",
-		"rimraf": "^3.0.2"
-	}
+  "private": true,
+  "scripts": {
+    "clean": "rimraf output && rimraf .pulp-cache",
+    "build": "eslint src && pulp build -- --censor-lib --strict",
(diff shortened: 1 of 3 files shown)


JordanMartinez:
Just curious, but is there any precedence for these functions? And could you provide an example of how you've used them?

mikesol:
> Just curious, but is there any precedence for these functions? And could you provide an example of how you've used them?

I can't find any precedence, but I've used the logic from `cayo` when working with lenses and wanting to make sure a law holds. I also used `moyo` when coming up with a set of laws for `IsEvent` from event.

Both functions turn the use of identity and mempty into a verb (a function) instead of a noun (the term), so they felt more laws-y to read.

The functions are trivial to write, so it's easy to implement them _ad hoc_. The way I'm framing a discussion is, given how trivial and self-evident they are, is there something "natural" about them that would only make sense in a prelude? Or is there another lib that'd be a better fit? Or are they too trivial/limited to go in any library & they should be hand-rolled when needed?

JordanMartinez:
I think I'm against adding them here for the following reasons:
- no precedence
- it's not clear to me that multiple people benefit from these being defined here
- it's not clear to me whether there may be other alternatives to using `cayo`/`mayo` that could also address the need you have for using them.

mikesol:
👍 I'll close the PR.

PR #296, opened 2022-06-07
Use `var` instead of `const`

triallax:
**Description of the change**

Replace `const` with `var`, since esbuild cannot lower `const` and `let` to `var` at the moment, preventing transpiling to ES5.

---

**Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Breaking changes:
 New features:
 
 Bugfixes:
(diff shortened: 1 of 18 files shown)


natefaubion:
AFAIK, the ecosystem in general has moved over to const with the 0.15 release. While esbuild may not support lowering these to ES5, many other bundlers and plugins do. I personally don't think that we should retroactively make this change across the ecosystem.

triallax:
@natefaubion that's fair.

PR #299, opened 2022-08-16
Remove array allocation and local array ffi from the Show instance for records

ajnsit:
**Description of the change**

Removes array allocation from the implementation of show for records, and also consequently, removes the need for local ffi implementations of array functions just to show records.

---

**Checklist:**

- [X] Added the change to the changelog's "Unreleased" section with a reference to this PR (e.g. "- Made a change (#0000)")
- [X] Added a test for the contribution (if applicable)

### DIFF
--- a/src/Data/Show.purs
+++ b/src/Data/Show.purs
@@ -5,7 +5,10 @@ module Data.Show
   , showRecordFields
   ) where
 
+import Data.Semigroup ((<>))
 import Data.Symbol (class IsSymbol, reflectSymbol)
+import Data.Unit (Unit)
+import Data.Void (Void, absurd)
 import Prim.Row (class Nub)
 import Prim.RowList as RL
 import Record.Unsafe (unsafeGet)
@@ -19,6 +22,9 @@ import Type.Proxy (Proxy(..))
 -- | value as the expression `x`.
 class Show a where
   show :: a -> String
+
+instance showUnit :: Show Unit where
+  show _ = "unit"
 
 instance showBoolean :: Show Boolean where
   show true = "true"
@@ -42,6 +48,9 @@ instance showArray :: Show a => Show (Array a) where
 instance showProxy :: Show (Proxy a) where
   show _ = "Proxy"
 
+instance showVoid :: Show Void where
+  show = absurd
+
 instance showRecord ::
   ( Nub rs rs
   , RL.RowToList rs ls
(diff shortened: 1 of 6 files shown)


src/Data/Show.purs
```diff
@@ -42,32 +48,45 @@ instance showArray :: Show a => Show (Array a) where
...
   ( Nub rs rs
   , RL.RowToList rs ls
   , ShowRecordFields ls rs
   ) =>
   Show (Record rs) where
   show record = case showRecordFields (Proxy :: Proxy ls) record of
-    [] -> "{}"
-    fields -> intercalate " " [ "{", intercalate ", " fields, "}" ]
+    "" -> "{}"
+    r -> "{ " <> r <> " }"
```
JordanMartinez:
```purs
show record = "{" <> showRecordFields (Proxy :: Proxy ls) record) <> "}"
```

src/Data/Show.purs
```diff
@@ -42,32 +48,45 @@ instance showArray :: Show a => Show (Array a) where
...
-  showRecordFields _ _ = []
-
+  showRecordFields _ _ = ""
+else
+instance showRecordFieldsConsNil ::
+  ( IsSymbol key
+  , Show focus
+  ) =>
+  ShowRecordFields (RL.Cons key focus RL.Nil) row where
+  showRecordFields _ record = key <> ": " <> show focus
```
JordanMartinez:
```purs
" " <> key <> ": " <> show focus <> " "
```

src/Data/Show.purs
```diff
@@ -42,32 +48,45 @@ instance showArray :: Show a => Show (Array a) where
...
+    focus = unsafeGet key record :: focus
+else
 instance showRecordFieldsCons ::
   ( IsSymbol key
   , ShowRecordFields rowlistTail row
   , Show focus
   ) =>
   ShowRecordFields (RL.Cons key focus rowlistTail) row where
-  showRecordFields _ record = cons (intercalate ": " [ key, show focus ]) tail
+  showRecordFields _ record = key <> ": " <> show focus <> ", " <> tail
```
JordanMartinez:
```purs
" " <> key <> ": " <> show focus <> "," <> tail
```

JordanMartinez (commented):
Correct me if I'm wrong, but can't this code...
```purs
show record = case showRecordFields (Proxy :: Proxy ls) record of
  "" -> "{}"
  r -> "{ " <> r <> " }"
```
be simplified to just
```purs
show record = "{" <> showRecordFields (Proxy :: Proxy ls) record) <> "}"
```
if the implementation is updated to the following?

src/Data/Show.purs
ajnsit: You are right, that does work and is a simpler way to accomplish the same thing. I have made the changes

garyb (approved):
Nice :+1:

JordanMartinez (approved):

src/Data/Show.purs
```diff
@@ -20,6 +23,9 @@ import Type.Proxy (Proxy(..))
 class Show a where
   show :: a -> String
 
+instance showUnit :: Show Unit where
+  show _ = "unit"
+
```
JordanMartinez, line 28: Could you clarify why this change was also made? And I guess the question is, should it be made?

garyb: Same question for `Void` actually - maybe an earlier iteration of this change needed these in scope for its implementation?

Since all this is in the same library I guess it doesn't matter too much, but the general principle I used years ago when we split the `Prelude` into modules, was that only instances for `Prim` types or those that were unavoidably depended upon are introduced in the class module, otherwise they were introduced with the type.

ajnsit: This was needed to avoid cycles in the import tree. I needed the semigroup instance for strings inside the show module, but `Data.Semigroup` depends on `Data.Void` and `Data.Unit`, which in turn depended on `Data.Show` for defining their show instance.

The rationale for why I resolved the cyclic imports this way is - by removing typeclass dependencies from "core" data structures like `Unit` and `Void`, we can trivially ensure they never cause cyclical imports and any typeclasses can always import the data structures. This way we only need to organise and resolve potential cyclical imports between typeclass modules like `Data.Show` and `Data.Semigroup`.

JordanMartinez: Gotcha. Thanks for clarifying!

PR #302, opened 2022-10-30
Guide type inference for Record type class instances via newtypes, not Proxies

JordanMartinez:
**Description of the change**

While working on [purescript-json-codecs](https://github.com/JordanMartinez/purescript-json-codecs), I realized that all these proxy args used to guide type inference could be replaced with newtypes now that we have polykinds and explicit kind signatures. Doing so removes one or two curried functions from the resulting code at the cost of some readability.

Naming used was the first letter of the type class name (e.g. `S` for `Semigroup`) followed by `Record`. If two newtypes were needed, then `1` or `2` was thrown in there somewhere.

Throughout this code, I only really use 2-3 newtypes. I wonder if these newtypes should actually be defined somewhere so one can reuse them to implement Record instances for various type classes.

Regardless, if this is accepted, I think the newtypes' names should be changed to prevent users from accidentally importing them since they really are intended to be internal. Perhaps something like `__SRecord1`?

---

**Checklist:**

### DIFF
--- a/src/Data/Bounded.purs
+++ b/src/Data/Bounded.purs
@@ -3,6 +3,7 @@ module Data.Bounded
   , bottom
   , top
   , module Data.Ord
+  , BRecord
   , class BoundedRecord
   , bottomRecord
   , topRecord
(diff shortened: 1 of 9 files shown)


JordanMartinez:
Closing due to no activity here.

PR #313, opened 2025-01-31
Add lazily evaluated versions of the conditional functions for applicatives and monads

JoelLefkowitz:
`ifM`, `whenM` and `unlessM` are really helpful for handling monads expressively. However, they require both of the possible outputs to be constructed. This can be problematic when the Monad is `Effect`, as PureScript is strictly evaluated, since both the side effects will still need to be constructed, even if they are not launched. This is because they may be expensive to compute especially if they use the imported JS functions.

Here is an example where a lazily evaluated version `ifM'` would be helpful:

```purs
create :: User -> Effect String
update :: User -> Effect String
exists :: User -> Effect Boolean

main :: Effect Unit
main = do
  response <- ifM' exists update create user
  log response
```

This PR creates lazy versions of each of the conditional functions for Applicatives and Monads:

- `ifM'`
- `unless'`
- `unlessM'`
- `when'`
- `whenM'`

This is following the same approach as `maybe'`:

> Similar to `maybe` but for use in cases where the default value may be
> expensive to compute. As PureScript is not lazy, the standard `maybe` has
> to evaluate the default value before returning the result, whereas here
> the value is only computed when the `Maybe` is known to be `Nothing`.

```purs
maybe' :: forall a b. (Unit -> b) -> (a -> b) -> Maybe a -> b
maybe' g _ Nothing = g unit
maybe' _ f (Just a) = f a
```

---

**Checklist:**

### DIFF
--- a/src/Control/Bind.purs
+++ b/src/Control/Bind.purs
@@ -12,6 +12,7 @@ module Control.Bind
   , composeKleisliFlipped
   , (<=<)
   , ifM
+  , ifM'
   , module Data.Functor
   , module Control.Apply
   , module Control.Applicative
@@ -148,3 +149,23 @@ infixr 1 composeKleisliFlipped as <=<
 -- | ```
 ifM :: forall a m. Bind m => m Boolean -> m a -> m a -> m a
 ifM cond t f = cond >>= \cond' -> if cond' then t else f
+
+-- | Similar to `ifM` but for use in cases where one of the monadic actions may
+-- | be expensive to compute or be responsible for side effects. As PureScript
+-- | is not lazy, the standard `ifM` has to construct both monadic actions
+-- | before returning the result, whereas here only the corresponding monadic
+-- | action is constructed.
+-- |
+-- | ```purescript
+-- | main :: Effect Unit
+-- | main = do
+-- |   response <- ifM' exists update create user
+-- |   log response
+-- |
+-- |   where
+-- |     create :: User -> Effect String
+-- |     update :: User -> Effect String
+-- |     exists :: User -> Effect Boolean
+-- | ```
+ifM' :: forall a b m. Bind m => (a -> m Boolean) -> (a -> m b) -> (a -> m b) -> a -> m b
+ifM' cond t f a = cond a >>= \cond' -> if cond' then t a else f a
--- a/src/Control/Bind.purs
+++ b/src/Control/Bind.purs
@@ -151,10 +151,9 @@ ifM :: forall a m. Bind m => m Boolean -> m a -> m a -> m a
 ifM cond t f = cond >>= \cond' -> if cond' then t else f
 
 -- | Similar to `ifM` but for use in cases where one of the monadic actions may
--- | be expensive to compute or be responsible for side effects. As PureScript
--- | is not lazy, the standard `ifM` has to construct both monadic actions
--- | before returning the result, whereas here only the corresponding monadic
--- | action is constructed.
+-- | be expensive to compute. As PureScript is not lazy, the standard `ifM` has
+-- | to construct both monadic actions before returning the result, whereas here
+-- | only the corresponding monadic action is constructed.
 -- |
 -- | ```purescript
 -- | main :: Effect Unit
--- a/src/Control/Bind.purs
+++ b/src/Control/Bind.purs
@@ -151,9 +151,7 @@ ifM :: forall a m. Bind m => m Boolean -> m a -> m a -> m a
 ifM cond t f = cond >>= \cond' -> if cond' then t else f
 
 -- | Similar to `ifM` but for use in cases where one of the monadic actions may
--- | be expensive to compute. As PureScript is not lazy, the standard `ifM` has
--- | to construct both monadic actions before returning the result, whereas here
--- | only the corresponding monadic action is constructed.
+-- | be expensive to construct.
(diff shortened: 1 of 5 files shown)


garyb:
> This is problematic when the Monad is `Effect`, as PureScript is strictly evaluated, since both the side effects will still happen.

Unless the `Effect` has been implemented improperly via the FFI it won't be evaluated unconditionally by `ifM`, etc - `Effect`s are essentially thunks, so side effects only occur on `bind`. 

One potential argument for adding these would be if it's expensive to construct the `Effect` even without evaluating it, I'm not sure I've ran into that scenario exactly, but another possible solution there without adding new functions would be to do something like `?cond >>= if _ then ?eff1 else ?eff2`.

JoelLefkowitz:
@garyb you're right the side effects are not evaluated and I had a bad foreign import. I've changed the PR content to focus on the issue of effects that are expensive to compute. 

An example I've run into where the effects are expensive to construct is when using a JS logging framework:

```purs
foreign import log :: String -> Effect Unit

main :: Effect Unit
main = do
    ifM (...) (...) (log "...")
```

Another small advantage of the function is that it can be more ergonomic to write `ifM' exists update create user`
rather than `ifM (exists user) (update user) (create user)`

JoelLefkowitz:
@garyb could you please take another look at this?

src/Control/Bind.purs
```diff
@@ -148,3 +149,22 @@ infixr 1 composeKleisliFlipped as <=<
 -- | ```
 ifM :: forall a m. Bind m => m Boolean -> m a -> m a -> m a
 ifM cond t f = cond >>= \cond' -> if cond' then t else f
+
+-- | Similar to `ifM` but for use in cases where one of the monadic actions may
+-- | be expensive to compute. As PureScript is not lazy, the standard `ifM` has
```
garyb: Perhaps "construct" rather than "compute" in this first sentence? And really I think that would suffice for the explanation, elaborating further on strict eval semantics only really applies to people who are newly coming to PS from Haskell. :wink:

Having the code example is still good though :+1:

src/Control/Monad.purs
```diff
@@ -55,16 +60,22 @@ liftM1 f a = do
...
 whenM :: forall m. Monad m => m Boolean -> m Unit -> m Unit
-whenM mb m = do
-  b <- mb
-  when b m
+whenM mb m = ifM mb m $ pure unit
+
+-- | Perform a monadic action lazily when a condition is true, where the conditional
+-- | value is also in a monadic context.
+whenM' :: forall m a. Monad m => (a -> m Boolean) -> (a -> m Unit) -> a -> m Unit
+whenM' mb m = ifM' mb m $ \_ -> pure unit
```
garyb:
```suggestion
whenM' mb m = ifM' mb m \_ -> pure unit
```

src/Control/Monad.purs
```diff
@@ -55,16 +60,22 @@ liftM1 f a = do
...
 unlessM :: forall m. Monad m => m Boolean -> m Unit -> m Unit
-unlessM mb m = do
-  b <- mb
-  unless b m
+unlessM mb = whenM $ not <$> mb
+
+-- | Perform a monadic action lazily unless a condition is true, where the conditional
+-- | value is also in a monadic context.
+unlessM' :: forall m a. Monad m => (a -> m Boolean) -> (a -> m Unit) -> a -> m Unit
+unlessM' mb = whenM' $ \x -> mb x >>= not >>> pure
```
garyb:
```suggestion
unlessM' mb = whenM' \x -> mb x >>= not >>> pure
```

JoelLefkowitz:
Thanks @garyb I've cut down the `ifM'` docstring and removed the superfluous `$` uses

garyb:
I just realised all the other comments refer to performing an action lazily, could you update those too? (Sorry)

Also there's an unused dependency that is causing the build failure just now. 😉

JoelLefkowitz:
Thanks @garyb no problem :)

I've reworded them from `Perform a monadic action lazily unless a condition` to `Construct a monadic action unless` etc. Is that what you had in mind?

JoelLefkowitz:
Hey @garyb could this please be merged?

garyb:
Sorry sorry, I've been crazy busy lately. Perhaps "Construct and evaluate" since it sounds like now it's just constructing them but not doing anything with them?

JoelLefkowitz:
Thanks @garyb no worries :)

I've reworded it like this:

```purs
-- | Perform a monadic action when a condition is true, where the conditional
-- | value is also in a monadic context.
whenM :: forall m. Monad m => m Boolean -> m Unit -> m Unit
whenM mb m = ifM mb m $ pure unit

-- | Perform a monadic action when a condition is true, without constructing it
-- | otherwise, where the conditional value is also in a monadic context.
whenM' :: forall m a. Monad m => (a -> m Boolean) -> (a -> m Unit) -> a -> m Unit
whenM' mb m = ifM' mb m \_ -> pure unit
```

I think it makes it a lot clearer that they both will only evaluate the monadic action when the condition is true, but only `whenM` will always construct it.

JoelLefkowitz:
Hey @garyb so sorry to tag you again in this since you are busy with other things but I'd really like to complete this PR, do you have a moment to approve it?

PR #314, opened 2025-02-10
Avoid `RangeError` in `arrayBind` foreign implementation

### LINKED ISSUE #309: `arrayBind` causes Maximum call stack size exceeded

pete-murphy:
**Description of the change**

Fixes #309. See the [related discussion on Discourse](https://discourse.purescript.org/t/stack-safe-cartesian-product/4841).

- ec814e4c4bbc7cf43387b9fc0657a3023c069e47 adds a failing test, reproducing the issue by passing 106,000 arguments to `Array.prototype.push.apply`
- a7159c7e56cc67cd17168c439c01e661bc32bada adds a case to use `flatMap` if available in the `arrayBind` implementation
- e9ad550295918e59ba5e4e2f670ba2e5724e068f simplifies the other case, and makes it stack-safe but gives up the `Array.prototype.push.apply` optimization

I don't know the motivation behind that optimization, so I am likely not aware of the tradeoffs. I tried [some benchmarks](https://observablehq.com/d/20091fbb067cb04f) in various JS run times, not seeing wildly-different results between existing solution (`#ff725c`), `flatMap` (`#4269d0`), and simplified version added in e9ad550295918e59ba5e4e2f670ba2e5724e068f (`#efb118`):

<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://github.com/user-attachments/assets/f961d1ff-2ffa-442e-97f0-0d8fac63bc6e">
  Multiple line charts showing benchmark results for three alternative implementations of arrayBind in four different JS runtimes: Safari, Firefox, Node, and Chrome
</picture>

See also #311 

---

**Checklist:**

### DIFF
--- a/src/Control/Bind.js
+++ b/src/Control/Bind.js
@@ -1,18 +1,21 @@
-export const arrayBind = function (arr) {
-  if (typeof Array.prototype.flatMap === "function") {
-    return function (f) {
-      return arr.flatMap(f);
-    };
-  }
-
-  return function (f) {
-    const result = [];
-    for (let i = 0, l = arr.length; i < l; i++) {
-      const xs = f(arr[i]);
-      for (let j = 0, m = xs.length; j < m; j++) {
-        result.push(xs[j]);
-      }
+export const arrayBind =
+  typeof Array.prototype.flatMap === "function"
+    ? function (arr) {
+      return function (f) {
+        return arr.flatMap(f);
+      };
     }
-    return result;
-  };
-};
+    : function (arr) {
+      return function (f) {
+        var result = [];
+        var l = arr.length;
+        for (var i = 0; i < l; i++) {
+          var xs = f(arr[i]);
+          var k = xs.length;
+          for (var j = 0; j < k; j++) {
+            result.push(xs[j]);
+          }
+        }
+        return result;
+      };
+    };
--- a/src/Control/Bind.js
+++ b/src/Control/Bind.js
@@ -1,4 +1,10 @@
 export const arrayBind = function (arr) {
+  if (typeof Array.prototype.flatMap === "function") {
+    return function (f) {
+      return arr.flatMap(f);
+    };
+  }
+
   return function (f) {
     var result = [];
     for (var i = 0, l = arr.length; i < l; i++) {
--- a/src/Control/Bind.js
+++ b/src/Control/Bind.js
@@ -6,9 +6,12 @@ export const arrayBind = function (arr) {
   }
 
   return function (f) {
-    var result = [];
-    for (var i = 0, l = arr.length; i < l; i++) {
-      Array.prototype.push.apply(result, f(arr[i]));
+    const result = [];
+    for (let i = 0, l = arr.length; i < l; i++) {
+      const xs = f(arr[i]);
+      for (let j = 0, m = xs.length; j < m; j++) {
+        result.push(xs[j]);
+      }
     }
     return result;
   };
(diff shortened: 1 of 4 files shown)


src/Control/Bind.js
```diff
@@ -1,8 +1,17 @@
 export const arrayBind = function (arr) {
+  if (typeof Array.prototype.flatMap === "function") {
```
natefaubion: Do we need to check this on every bind invocation? I think this could probably be a static check.

src/Control/Bind.js
```diff
@@ -1,8 +1,17 @@
...
   return function (f) {
-    var result = [];
-    for (var i = 0, l = arr.length; i < l; i++) {
-      Array.prototype.push.apply(result, f(arr[i]));
+    const result = [];
+    for (let i = 0, l = arr.length; i < l; i++) {
+      const xs = f(arr[i]);
+      for (let j = 0, m = xs.length; j < m; j++) {
+        result.push(xs[j]);
+      }
```
natefaubion: I don't think anything in prelude ffi uses const or for/let. We should probably continue to use `var` for consistency for now.

natefaubion (changes requested):

src/Control/Bind.js
pete-murphy: Updated here https://github.com/purescript/purescript-prelude/pull/314/commits/f8783251d38b8d2902397f0f481ccfe8b760a779, but not sure if it's what you had in mind?

src/Control/Bind.js
pete-murphy: Changed to use `var` and pulled the length variable out to be more consistent with other FFI code in this repo.

src/Control/Bind.js
```diff
@@ -1,9 +1,21 @@
...
+        for (var i = 0; i < l; i++) {
+          const xs = f(arr[i]);
+          const k = xs.length;
+          for (var j = 0; j < k; j++) {
+            result.push(xs[j]);
+          }
+        }
+        return result;
+      };
+    };
```
pete-murphy: Maybe worth pulling this out to a standalone function, to export and test

natefaubion: I don't think a legacy FFI implementation should be a public Prelude export. If we want to test it we should just duplicate it, for now, and maybe add a comment.

src/Control/Bind.js
natefaubion: I don't think this made it in. I still see let/const.

pete-murphy: Force-pushed, `const` only appears in top-level declaration

natefaubion (approved):

garyb (approved):

PR #316, opened 2025-05-08
Remove NaN tests for Ord and Eq

### LINKED ISSUE #306: `ordNumber` is not compatible with IEEE 754 floats
`purs` compiles the following:

```
(1.0 < nan) == false
```
to

```
1.0 < nan === false
```

but if we instead go through the `Ord` API we can observe different behavior that is not compatible with IEEE 754:

```
compare 1.0 nan == GT
```

In addition to just being wrong, the more practical issue is that any alternative backend that does not inline the use of comparison operators to operators that implement IEEE 754 correctly won't pass the prelude test suite. For example `backend-es` does not pass the prelude test suite.

One simple fix that was floated around in Discord is to just remove the tests using `NaN` from the test suite and leave the behavior undefined.

anttih:
These tests rely on the `purs` inliner to provide correct behavior. With this change we leave the `Ord` and `Eq` `NaN` behavior undefined. A backend can choose how comparisons with `NaN` work.

Fixes #306

### DIFF
--- a/test/Test/Main.purs
+++ b/test/Test/Main.purs
@@ -32,9 +32,6 @@ testOrd x y ord =
         ("(compare " <> show x <> " " <> show y <> " ) is not equal to " <> show ord)
         $ (compare x y) == ord
 
-nan :: Number
-nan = 0.0/0.0
-
 plusInfinity :: Number
 plusInfinity = 1.0/0.0
 
@@ -43,8 +40,6 @@ minusInfinity = -1.0/0.0
 
 testOrderings :: AlmostEff
 testOrderings = do
-    assert "NaN shouldn't be equal to itself" $ nan /= nan
-    assert "NaN shouldn't be equal to itself" $ (compare nan nan) /= EQ
     testOrd 1.0    2.0 LT
     testOrd 2.0    1.0 GT
     testOrd 1.0    (-2.0) GT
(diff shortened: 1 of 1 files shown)


garyb:
I don't know if this is right exactly, these NaN operations are part of IEEE-754 which we probably use to make other assumptions about numbers too.

anttih:
@garyb Sure, but how do we solve #306? I thought it was agreed that we can consider `NaN` behavior as undefined for `Ord`, thus we could remove the tests. Do you have other suggestions?

anttih:
I don't think we do other assumptions with regards to `NaN`. If we do, then we at least don't have tests for those.

natefaubion:
While these are more problematic due to operator inlining, looks like there are some more Ord/Eq tests?

anttih:
I've now removed all `Ord` and `Eq` tests for `NaN`.

natefaubion (approved):
@garyb Right now the instances/inlining situation is just incoherent. These tests aren't testing IEEE conformance so much as they are testing that the compiler inlines these operators. For example, use something other than prelude comparison operators, and these won't work!

I think if we have these tests anywhere they would need to be in the compiler with the JS backend because that's the only place this might make sense to validate.

garyb:
Yeah, that's all fine, I'd completely forgotten about the previous discussion @anttih linked even though I participated. :smile:

garyb (approved):

anttih:
Can someone merge this?

